PR #10, opened 2020-07-23
Adding the ValidateChip pipeline

gbggrant:


pipelines/broad/arrays/validate_chip/ValidateChip.dev.options.json
```diff
@@ -0,0 +1,9 @@
+{
+  "read_from_cache": true,
```
jessicaway, line 2: Do we still need the 3 different options files?

gbggrant: I'm not sure.  Can't remember what google_project is doing there anyways.  I'll see what happens.

gbggrant: Looks like they're not needed.  Have updated.

jessicaway: 👍

PR #20, opened 2020-07-31
GL-1114 copied over skylab atac pipelines to warp

nikellepetrillo:

PR #24, opened 2020-08-03
Np gl-1114 update atac workflows docs and dockers

nikellepetrillo:


kbergin:
Does this PR also include everything in PR #20 ? Seems like it from first glance.

nikellepetrillo:
@kbergin yes, i believe it is because I branched off of https://github.com/broadinstitute/warp/pull/20 when creating this branch

PR #26, opened 2020-08-03
Copy over MultiSampleArrays pipeline

sophiacrennan:


pipelines/broad/arrays/multi_sample/build_docker.sh
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash
```
gbggrant, line 1: I think the docker files should go into a subdirectory of dockers/broad/

sophiacrennan:
test_optimus_snrna is failing because permission is denied to access the gotc cromwell monitioring script. A ticket was made to fix that issue.

PR #35, opened 2020-08-13
Get release script working

jessicaway:
change references to dsde-pipelines to make release scripts work in warp


scripts/build_workflow_release.sh
```diff
@@ -2,7 +2,7 @@
 set -e -o pipefail
 
 declare -r SCRIPT_DIR=$(cd $(dirname $0) && pwd)
-declare -r DSDE_PIPELINES_ROOT=$(cd $(dirname ${SCRIPT_DIR}) && pwd)
+declare -r WARP_ROOT=$(cd $(dirname ${SCRIPT_DIR}) && pwd)
```
pshapiro4broad, line 5: You could name it `REPO_ROOT` in case we move everything again.

scripts/get_changed_pipeline_worklow_test_args.sh
```diff
@@ -22,34 +22,34 @@ function pipeline_to_args() {
...
     MultiSampleArrays)
       echo Arrays -a Multi ${common_args};;
-    ExomeGermlineSingleSample)
-      echo GermlineSingleSample -d Exome ${common_args};;
-    ExomeReprocessing)
-      echo Reprocessing -d Exome ${common_args};;
-    JointGenotyping)
-      echo JointGenotyping -d Exome ${common_args};
-      echo JointGenotyping -d WGS --env ${env} -t Plumbing -b develop ${uncached};;
+    #ExomeGermlineSingleSample)
```
pshapiro4broad, line 25: If these are commented out, then an unsupported test will result in empty output. Is that what we want to do? Vs generating an error or printing a message, etc.

scripts/release_pipeline_to_github.sh
```diff
@@ -8,6 +8,7 @@
 # as an environment variable
 set -e
 
+declare -r REPO="broadinstitute/warp"
```
pshapiro4broad, line 11: Might not be worth it, but there is probably a way to derive this using git commands instead of hardcoding it.

scripts/get_changed_pipeline_worklow_test_args.sh
jessicaway: My thought was that we would bring them back in as the pipelines were brought over. We can do it a different way if you prefer

scripts/release_pipeline_to_github.sh
jessicaway: Looks like the following command will do it: 
`REPO=$(git remote -v | grep "fetch" | sed 's/git@github.com://; s/origin//; s/.git//; s/(fetch)//')`
Do we like that better?  (I tried to reduce the sed commands, but could not get it to work properly)

scripts/build_workflow_release.sh
jessicaway: I will change this

scripts/release_pipeline_to_github.sh
pshapiro4broad: I think that's better. I got it to work with a slightly simpler `sed:
```
REPO=$(git remote -v | grep fetch | sed -E 's/.*git@github.com:(.*).git.*/\1/')
```

scripts/get_changed_pipeline_worklow_test_args.sh
pshapiro4broad: It's OK, I guess we'll notice if things do nothing vs forcing it to fail.

scripts/release_pipeline_to_github.sh
jessicaway: Thanks!

PR #48, opened 2020-08-25
Releasing SS2 and Optimus for HCA update 

khajoue2:


jessicaway:
Smart tests, are not fully hooked up yet, this failure is OK to ignore for now.

beta-pipelines/skylab/ATAC/ATAC.wdl
```diff
@@ -50,6 +52,7 @@ workflow ATAC {
     min_map_quality: "the minimum mapping quality to be filtered by samtools view and snap-pre (snaptools task)"
     max_fragment_length: "the maximum fragment length for filtering out reads by gatk and snap-pre (snaptools task)"
     output_base_name: "base name to be used for the pipelines output and intermediate files"
+    bin_size_list: "space separated list of bins to generate"
```
kbergin, line 55: I'm not clear on what "space separated list" means. Isn't this just a numerical value regarding the number of bins? No reason to block this merge to master on this comment but maybe we can follow up on it. @ekiernan @sophiacrennan

dockers/skylab/loom-output/create_loom_optimus.py
```diff
@@ -430,12 +434,30 @@ def main():
...
+    parser.add_argument(
+        "--input_name",
+        dest="input_name",
+        help= "sequencing_input.biomaterial_core.biomaterial_id in HCA metadata, defined by the user",
+    )
+
+    parser.add_argument(
+        "--input_id_metadata_field",
+        dest="input_id_metadata_field",
+        help= "sequencing_process.provenance.document_id: [UUID] defined by the user",
```
kbergin, line 452: If it's a UUID it's not really defined by the user for HCA, but maybe you mean defined by user when the pipeline is run? I like that this documentation notes what field is used in HCA given these fields are basically entirely for HCA, but I wonder if they should also give other users more context? Like for input_id_metadata_field it could also say "the name of the metadata field used for the input_id"

Is this saying that the way it will appear in the outputs is input_id_metadata-field: [UUID from input_id] ? 

Like say the `input_id_metadata_field` is called `my_favorite_metadata` and the UUID in `input_id` was `ABC123`

Would the loom say `my_favorite_metadata:ABC123` ?

tests/skylab/smartseq2_multisample/pr/ValidateMultiSampleSmartSeq2.wdl
```diff
@@ -18,7 +18,7 @@ task ValidateSmartSeq2Plate {
   >>>
   
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-loom-output:0.0.2"
+    docker: "quay.io/humancellatlas/secondary-analysis-loom-output:0.0.3-fk-2"
```
kbergin, line 21: This seems like a very unofficial name - should this have the initials in it? Why isn't it a released version of the docker image?

beta-pipelines/skylab/ATAC/ATAC.wdl
ekiernan: In the README we say: "List of bin sizes used to generate cell-by-bin matrices"- is that clear enough? We confirmed that this option can be a list of values, not just a single value.

kbergin: That's super interesting, I didn't realize it could be a list! I think the meta is ok as it is then, although I do think the README wording is a bit clearer.

PR #60, opened 2020-09-02
copying over WholeGenomeGermline/ExomeGermline wdls and structs

nikellepetrillo:


pipelines/broad/germline/single_sample/exome/test_inputs/load_samples.tsv
```diff
@@ -0,0 +1,501 @@
+HCVJNBBXX.7	AE68915CK
```
jessicaway, line 1: We probably don't need this file

pipelines/broad/germline/single_sample/exome/test_inputs/load_template.json
```diff
@@ -0,0 +1,56 @@
+  {
```
jessicaway, line 1: Don't need this either

PR #88, opened 2020-09-30
changed the pipelines paths to the Fastqprocessing

kishorikonwar:


pipelines/skylab/optimus/Optimus.wdl
```diff
@@ -1,6 +1,6 @@
 version 1.0
 
-import "../../../FastqProcessing.wdl" as FastqProcessing
+import "../../../tasks/FastqProcessing.wdl" as FastqProcessing
```
kbergin, line 3: doesn't this need to be 
```suggestion
import "../../../tasks/skylab/FastqProcessing.wdl" as FastqProcessing
```
?

jessicaway: ➕

kbergin (changes requested):
The path still needs to be updated to the appropriate one, then it looks good.

kbergin (approved):

khajoue2 (approved):

PR #95, opened 2020-10-02
Adding a note about haplotype caller version

kbergin:
In this pipeline, by default the variant calling is done with GATK3.5 unless the user specifies an input parameter to use GATK4. We don't explain that anywhere and our example inputs use the defaults. 

I would suggest to put more documentation / example inputs but my hope is we will be going to DRAGEN-GATK genome pipeline soon and this will be deprecated so maybe not worth it. 

Other side question - I thought we updated documentation about the cromwell version support but at least develop seems to still note it was successfully tested on cromwell 47 - should we say it's now tested on cromwell 52 and newer? @ekiernan


ekiernan (approved):
Updated the Cromwell version note and made minor change to the wording.

jessicaway:
add to whitelist

jessicaway (approved):

PR #97, opened 2020-10-05
Run JG tests on PAPIv2

jessicaway:
Joint genotyping scientific tests have been hanging, likely due to the N2 machine type requested on PAPIv1. It has also been suggested that the JG pipelines are typically run on v2 in production, so switching tests to use v2 may be a more useful way to test


sophiacrennan (approved):
👍

nikellepetrillo (approved):

PR #105, opened 2020-10-13
releasing SS2 with loom bug fix

khajoue2:


kbergin (approved):
When the tests pass 👍

PR #132, opened 2020-11-02
Dev terra disk increase

gbggrant:
Increased memory to umi_tools and disk to cell_sort and gene_sort


khajoue2:
We fixed the issues with a fix to sctools. These changes are no longer needed

PR #137, opened 2020-11-05
GL-1271.  Simplified Input checking on SmartSeq2SingleSample

gbggrant:
Merged input checks for SmartSeq2SingleSample.wdl from the main wdl into HISAT2 tasks.
This in the interest of reducing the number of DRS calls, but also to reduce code complexity.
Actually, not really sure the -f checks added to the HISAT2 tasks are really adding any value.


sophiacrennan (approved):
Looks good!

khajoue2 (approved):

PR #146, opened 2020-11-23
Terra Scripts for MVP Optimus Processing

jessicaway:


pipelines/skylab/optimus/scripts/parse_terra.py
```diff
@@ -0,0 +1,121 @@
+import argparse
```
jessicaway, line 1: @kishorikonwar Is this the same script (more or less) as the "parse_terra_tsv.py" script that is already checked in? Let's not have the same script checked in twice please. Also I think the other one is more up-to-date, can we please make sure the workflow-controller calls the right one?

pipelines/skylab/optimus/scripts/fiss.ipynb
```diff
@@ -0,0 +1,307 @@
+{
```
jessicaway, line 1: Is this file needed? If so is it ready to be checked-in?

pipelines/skylab/optimus/scripts/json_templates.py
```diff
@@ -0,0 +1,30 @@
+optimus_inputs = {
```
jessicaway, line 1: Do we need this file?

pipelines/skylab/optimus/scripts/workspace_controller.py
```diff
@@ -0,0 +1,316 @@
+import argparse
```
jessicaway, line 1: We will put a ticket in the backlog to clean this code up at some point

nikellepetrillo (approved):

kishorikonwar (approved):

PR #149, opened 2020-11-24
Add "google_legacy_machine_selection" support

gbggrant:
Add "google_legacy_machine_selection" support for WGS Germline workflow to get it working under PapiV2


sophiacrennan (approved):

jessicaway (approved):

jessicaway:
Do we need this in Whole genome reprocessing and external whole genome reprocessing as well?

gbggrant:
This is a good question.  I'm not sure where External WGS gets its options - quite likely it's using this one.

jessicaway:
each pipeline has a separate one next to it, but I am not sure which one is being used for production processing

PR #170, opened 2020-12-15
Fix bash if space syntax

jennydaman:
Currently this error is produced

```
/cromwell_root/script: line 31: [-z: command not found
```

### DIFF
--- a/tasks/broad/Alignment.wdl
+++ b/tasks/broad/Alignment.wdl
@@ -54,7 +54,7 @@ task SamToFastqAndBwaMemAndMba {
     set -o pipefail
     set -e
 
-    if [-z ${BWA_VERSION}]; then
+    if [ -z ${BWA_VERSION}]; then
         exit 1;
     fi
 
--- a/tasks/broad/Alignment.wdl
+++ b/tasks/broad/Alignment.wdl
@@ -54,7 +54,7 @@ task SamToFastqAndBwaMemAndMba {
     set -o pipefail
     set -e
 
-    if [ -z ${BWA_VERSION}]; then
+    if [ -z ${BWA_VERSION} ]; then
         exit 1;
     fi
 


tasks/broad/Alignment.wdl
```diff
@@ -54,7 +54,7 @@ task SamToFastqAndBwaMemAndMba {
     set -o pipefail
     set -e
 
-    if [-z ${BWA_VERSION}]; then
+    if [ -z ${BWA_VERSION}]; then
```
gbggrant, line 57: Might as well put a space on both sides 
```suggestion
    if [ -z ${BWA_VERSION} ]; then
```
I have also read that the double bracket syntax is supposed to be better. '[[' vs '['

gbggrant (approved):
Thank you!

jessicaway (approved):

tasks/broad/Alignment.wdl
jennydaman: Thanks for catching that! `}]` is still invalid syntax, I made this PR in a hurry.

`[[` is a superset of `[`. I don't think it's necessary.

jessicaway:
retest this please

jessicaway:
ok to test

jessicaway:
@jennydaman, thank you for contributing this PR. Unfortunately we do not yet have the ability to safely test PRs from users outside our team. This is an issue we plan to address as soon as possible, but in this case I have created [a separate PR](https://github.com/broadinstitute/warp/pull/178) with your changes to allow for testing

jennydaman:
I've noticed, though the routine seems excessive for insertion of two whitespace characters.

jessicaway:
Yes, we are still working on balancing thorough testing with faster iteration time. It can be especially painful for small changes like this. Hopefully we will be able to come up with a simpler workflow for these small changes soon

jessicaway:
Replaced by [this PR ](https://github.com/broadinstitute/warp/pull/178)

PR #174, opened 2020-12-16
Kd add gdc pipeline

kbergin:
I've copied over the code from https://github.com/broadinstitute/gdc-wgs by @dheiman. The only modifications I've made were to rename the workflow to better match WARP naming schema and to add a pipeline version string. 

I've uploaded an inputs.json used in Terra to test this workflow, but replaced the reference paths to the matching google public references gs path. 

Most of this is a placeholder to get it into WARP so we can begin making edits to the workflow.


kbergin:
Need to update the example inputs json to use the appropriate reference, it's just pointed to a standard hg38 for now.

jessicaway:
add to whitelist

jessicaway (approved):

gbggrant (approved):

PR #175, opened 2020-12-16
Remove beta-pipelines from miniwdl check

jessicaway:


gbggrant (approved):
We should remember (add to testing Jira ticket) to put this back.

nikellepetrillo (approved):

PR #176, opened 2020-12-17
Remove custom entrypoint Picard Docker image

kbergin:
Picard:2.18.11 docker image has a custom entrypoint which Terra cromwell just ignores, but older versions of Cromwell cannot support. So that we can run this pipeline on GOTC which is Cromwell52 quickly, we are changing to a new docker image without a custom entrypoint.


gbggrant (approved):
Looks good - have you searched for other usages of picard-cloud?

kbergin:
I hadn't, but I did just now and this task is the only one that calls picard-cloud. Thanks!

sophiacrennan (approved):

PR #181, opened 2020-12-22
Allow Optimus to optionally count reads in stranded mode

sophiacrennan:


kishorikonwar (commented):
We need to make sure what the use_strand_info is used for

pipelines/skylab/optimus/Optimus.wdl
```diff
@@ -56,7 +59,7 @@ workflow Optimus {
 
   # version of this pipeline
 
-  String pipeline_version = "4.2.1"
+  String pipeline_version = "4.3.0"
```
kbergin, line 62: I wonder if this is a patch update since the input is optional and defaulted to have the same function as it did before?

pipelines/skylab/optimus/Optimus.changelog.md
```diff
@@ -1,3 +1,7 @@
+# 4.3.0
+
+* Added an optional input for the pipeline to read in stranded mode
```
kbergin, line 3: Let's note here that it defaults to false.

kishorikonwar (approved):

PR #188, opened 2021-01-05
GL-1345 ensure ordering is correct for combining optimus matrices

jessicaway:
Add a check to make sure that matrices being combined are in the same order


khajoue2 (approved):
lgtm

kishorikonwar (approved):

PR #141, opened 2020-11-06
Make reblocking default in single-sample workflows

ldgauthier:
Also refactor to move reblocking task

PR #197, opened 2021-01-07
Added DRAGEN-GATK features

michaelgatzen:
- Added WholeGenomeGermlineSingleSample.wdl with optional DRAGEN arguments to beta-pipelines/
    - Pipeline version -> 3.0.0
    - New workflow arguments
        - run_dragen_mode = false
        - perform_bqsr = true
        - Optional File dragmap_binary
        - Optional DragmapReference dragmap_reference
- Added new struct DragmapReference
- Changes to WholeGenomeGermlineSingleSample.wdl in pipelines/
    - Made output output_bqsr_reports optional as BQSR becomes optional
- Changes to several task WDLs
    - New file: Dragen.wdl
        - contains two tasks: ComposeSTRTableFile and CalibrateDragstrModel
    - New file: DragmapAlignment.wdl
        - Analog to Alignment.wdl for alignment with the open source version of the DRAGEN aligner
    - GermlineVariantDiscovery
        - HaplotypeCaller_GATK4_VCF
            - Arguments:
                - run_dragen_mode = false
                - Optional argument File dragstr_model, required when running dragen mode, however there is no check
                - gatk_docker made an optional argument with no default value. If not set and run_dragen_mode is false, it defaults to "us.gcr.io/broad-gatk/gatk:4.1.8.0". If run_dragen_mode is true, it is set to a docker containing the DRAGEN-GATK changes.
            - **Memory increased from 6.5 GiB to 10 GiB**
            - **maxRetries: 3**
    - Qc
        - ValidateSamFile
            - **Memory increased from 7 to 20 GiB**
            - **maxRetries: 3**
        - CollectRawWgsMetrics
            - **maxRetries: 3**
    - SplitLargeReadGroup
        - Included logic to choose between BWA mem and DragMap for alignment
        - Added (optional) respective arguments
    - UnmappedBamToAlignedBam
        - Workflow argument: perform_bqsr = true
        - (Optional) arguments for DragMap alignment
        - Included logic to choose between BWA mem and DragMap for alignment
        - Made BQSR and its output output_bqsr_reports optional
    - VariantCalling
        - Workflow arguments
            - run_dragen_mode = true
        - if running dragen mode, ComposeSTRTableFile and DragstrAutoCalibration is called
        - run_dragen_mode and dragstr_model passed on to HaplotypeCaller_GATK4_VCF
        - Task MergeBamouts
          - Command block using `<<< >>>` instead of `{ }`

### DIFF
--- a/tasks/broad/Dragen.wdl
+++ b/tasks/broad/Dragen.wdl
@@ -1,0 +1,102 @@
+version 1.0
+
+## Copyright Broad Institute, 2020
+##
+## This WDL defines tasks to use Dragen's DRAGstr approach to STR sequencing artifacts 
+## Indel genotype priors in the DRAGEN-Gatk pipeline. 
+##
+## Runtime parameters are often optimized for Broad's Google Cloud Platform implementation.
+## For program versions, see docker containers.
+##
+## LICENSING :
+## This script is released under the WDL source code license (BSD-3) (see LICENSE in
+## https://github.com/broadinstitute/wdl). Note however that the programs it calls may
+## be subject to different licenses. Users are responsible for checking that they are
+## authorized to run all programs before running this script. Please see the docker
+## page at https://hub.docker.com/r/broadinstitute/genomes-in-the-cloud/ for detailed
+## licensing information pertaining to the included programs.
+
+
+task ComposeSTRTableFile {
+    
+  input {
+    File ref_fasta
+    File ref_fasta_idx
+    File ref_dict
+    String docker = "us.gcr.io/broad-dsde-methods/broad-gatk-snapshots/dragen_final_test_v2"
+    Int preemptible_tries = 3
+  }
+
+  String base_name = sub(basename(ref_fasta), ".fa(sta)$","")
+  String str_table_file_name = base_name + ".str"
+  Int disk_size_gb = ceil(size([ref_fasta, ref_fasta_idx], "GiB")) + 
+                        21 # 1gb for the output file (usually far less than that) and 20 for the rest of the fs.
+
+  command <<<
+
+    gatk --java-options "-Xmx2g" \
+      ComposeSTRTableFile \
+        -R ~{ref_fasta} \
+        -O ~{str_table_file_name}
+
+  >>>
+
+  runtime {
+    docker: docker 
+    disks: "local-disk " + disk_size_gb + " HDD"
+    memory: "3 GiB"
+    preemptible: preemptible_tries
+  }
+
+  output {
+    File str_table_file = "~{str_table_file_name}"
+  }
+}
+
+task CalibrateDragstrModel {
+
+  input {
+    File ref_fasta
+    File ref_fasta_idx
+    File ref_dict
+    File str_table_file
+    File alignment ## can handle cram or bam.
+    String docker = "us.gcr.io/broad-dsde-methods/broad-gatk-snapshots/dragen_final_test_v2"
+    Int preemptible_tries = 3
+    Int threads = 4
+    Boolean use_ssd = true
+  }
+
+  String base_name = basename(alignment)
+  String out_file_name = base_name + ".dragstr"
+  Int disk_size_gb = ceil(size([ref_fasta, ref_fasta_idx, alignment, str_table_file], "GiB")) + 
+                        20 # 20 for the rest of the fs.
+
+  String parallel_args  = if (threads <= 1) then "" else "--parallel" 
+
+  command <<<
+    set -x
+    samtools index -@ ~{threads} ~{alignment}
+    gatk --java-options "-Xmx2g -DGATK_STACKTRACE_ON_USER_EXCEPTION=true" \
+      CalibrateDragstrModel \
+        -R ~{ref_fasta} \
+        -I ~{alignment} \
+        -str ~{str_table_file} \
+        -O ~{out_file_name} \
+        ~{parallel_args} \
+        --verbosity DEBUG
+
+  >>>
+
+  runtime {
+     docker: docker
+     disks: "local-disk " + disk_size_gb + (if use_ssd then " SSD" else " HDD")
+     memory: "3 GiB"
+     preemptible: preemptible_tries
+     cpu: threads
+  }
+
+  output {
+    File dragstr_model = "~{out_file_name}"
+  }
+}
--- a/tasks/broad/DragmapAlignment.wdl
+++ b/tasks/broad/DragmapAlignment.wdl
@@ -1,0 +1,109 @@
+version 1.0
+
+## Copyright Broad Institute, 2020
+##
+## This WDL defines tasks used for alignment of human whole-genome or exome sequencing data using Illumina's DRAGEN open source mapper.
+##
+## Runtime parameters are often optimized for Broad's Google Cloud Platform implementation.
+## For program versions, see docker containers.
+##
+## LICENSING :
+## This script is released under the WDL source code license (BSD-3) (see LICENSE in
+## https://github.com/broadinstitute/wdl). Note however that the programs it calls may
+## be subject to different licenses. Users are responsible for checking that they are
+## authorized to run all programs before running this script. Please see the docker
+## page at https://hub.docker.com/r/broadinstitute/genomes-in-the-cloud/ for detailed
+## licensing information pertaining to the included programs.
+
+import "../../structs/dna_seq/DNASeqStructs.wdl"
+
+# Read unmapped BAM, convert on-the-fly to FASTQ and stream to BWA MEM for alignment, then stream to MergeBamAlignment
+task SamToFastqAndDragmapAndMba {
+  input {
+    File input_bam
+    String output_bam_basename
+
+    ReferenceFasta reference_fasta
+    DragmapReference dragmap_reference
+
+    File picard_jar = "gs://broad-dsde-methods-mgatzen/dragen_evaluation/alignment/picard-2.23.7-SNAPSHOT-all.jar"
+    File dragmap_binary
+
+    Int compression_level
+    Int preemptible_tries
+    Boolean hard_clip_reads = false
+  }
+
+  Float unmapped_bam_size = size(input_bam, "GiB")
+  Float ref_size = size(reference_fasta.ref_fasta, "GiB") + size(reference_fasta.ref_fasta_index, "GiB") + size(reference_fasta.ref_dict, "GiB")
+  Float bwa_ref_size = ref_size + size(reference_fasta.ref_alt, "GiB") + size(reference_fasta.ref_amb, "GiB") + size(reference_fasta.ref_ann, "GiB") + size(reference_fasta.ref_bwt, "GiB") + size(reference_fasta.ref_pac, "GiB") + size(reference_fasta.ref_sa, "GiB")
+  Float dragmap_ref_size = size(dragmap_reference.reference_bin, "GiB") + size(dragmap_reference.reference_index_bin, "GiB") + size(dragmap_reference.hash_table_cmp, "GiB")
+  Float disk_multiplier = 6
+  Int disk_size = ceil(unmapped_bam_size + bwa_ref_size + dragmap_ref_size + (disk_multiplier * unmapped_bam_size) + 20)
+
+  String dragmap_commandline = "-1 reads1.fastq.gz -2 reads2.fastq.gz -r dragen_reference"
+
+  command <<<
+
+
+    set -euxo pipefail
+
+
+    chmod +x ~{dragmap_binary}
+
+    DRAGMAP_VERSION=$(~{dragmap_binary} --version)
+
+    if [ -z ${DRAGMAP_VERSION} ]; then
+        exit 1;
+    fi
+
+    mkdir dragen_reference
+    mv ~{dragmap_reference.reference_bin} ~{dragmap_reference.reference_index_bin} ~{dragmap_reference.hash_table_cmp} dragen_reference
+
+    samtools fastq -1 reads1.fastq.gz -2 reads2.fastq.gz -0 /dev/null -s /dev/null -n ~{input_bam}
+
+    ~{dragmap_binary} ~{dragmap_commandline} 2> >(tee ~{output_bam_basename}.dragmap.stderr.log >&2) | \
+    java -Dsamjdk.compression_level=~{compression_level} -Xms1000m -Xmx1000m -jar ~{picard_jar} \
+      MergeBamAlignment \
+      VALIDATION_STRINGENCY=SILENT \
+      EXPECTED_ORIENTATIONS=FR \
+      ATTRIBUTES_TO_RETAIN=X0 \
+      ATTRIBUTES_TO_REMOVE=RG \
+      ATTRIBUTES_TO_REMOVE=NM \
+      ATTRIBUTES_TO_REMOVE=MD \
+      ALIGNED_BAM=/dev/stdin \
+      UNMAPPED_BAM=~{input_bam} \
+      OUTPUT=~{output_bam_basename}.bam \
+      REFERENCE_SEQUENCE=~{reference_fasta.ref_fasta} \
+      PAIRED_RUN=true \
+      SORT_ORDER="unsorted" \
+      IS_BISULFITE_SEQUENCE=false \
+      ALIGNED_READS_ONLY=false \
+      CLIP_ADAPTERS=false \
+      ~{true='CLIP_OVERLAPPING_READS=true' false="" hard_clip_reads} \
+      ~{true='CLIP_OVERLAPPING_READS_OPERATOR=H' false="" hard_clip_reads} \
+      MAX_RECORDS_IN_RAM=2000000 \
+      ADD_MATE_CIGAR=true \
+      MAX_INSERTIONS_OR_DELETIONS=-1 \
+      PRIMARY_ALIGNMENT_STRATEGY=MostDistant \
+      PROGRAM_RECORD_ID="dragen-os" \
+      PROGRAM_GROUP_VERSION="${DRAGMAP_VERSION}" \
+      PROGRAM_GROUP_COMMAND_LINE="dragen-os ~{dragmap_commandline}" \
+      PROGRAM_GROUP_NAME="dragen-os" \
+      UNMAPPED_READ_STRATEGY=COPY_TO_TAG \
+      ALIGNER_PROPER_PAIR_FLAGS=true \
+      UNMAP_CONTAMINANT_READS=true \
+      ADD_PG_TAG_TO_READS=false
+  >>>
+  runtime {
+    docker: "michaelgatzen/dragen_os"
+    preemptible: preemptible_tries
+    memory: "40 GiB"
+    cpu: "16"
+    disks: "local-disk " + disk_size + " HDD"
+  }
+  output {
+    File output_bam = "~{output_bam_basename}.bam"
+    File bwa_stderr_log = "~{output_bam_basename}.dragmap.stderr.log"
+  }
+}
--- a/tasks/broad/GermlineVariantDiscovery.wdl
+++ b/tasks/broad/GermlineVariantDiscovery.wdl
@@ -92,8 +92,13 @@ task HaplotypeCaller_GATK4_VCF {
     Boolean make_bamout
     Int preemptible_tries
     Int hc_scatter
-    String gatk_docker = "us.gcr.io/broad-gatk/gatk:4.1.8.0"
+    Boolean run_dragen_mode = false
+    File? dragstr_model
+    String? gatk_docker
   }
+
+  String default_gatk_docker = if run_dragen_mode then "us.gcr.io/broad-dsde-methods/broad-gatk-snapshots/dragen_final_test_v2" else "us.gcr.io/broad-gatk/gatk:4.1.8.0"
+  String gatk_docker_to_use = select_first([gatk_docker, default_gatk_docker])
 
   String output_suffix = if make_gvcf then ".g.vcf.gz" else ".vcf.gz"
   String output_file_name = vcf_basename + output_suffix
@@ -119,6 +124,8 @@ task HaplotypeCaller_GATK4_VCF {
       -O ~{output_file_name} \
       -contamination ~{default=0 contamination} \
       -G StandardAnnotation -G StandardHCAnnotation ~{true="-G AS_StandardAnnotation" false="" make_gvcf} \
+      ~{true="--dragen-mode" false="" run_dragen_mode} \
+      ~{if defined(dragstr_model) then "--dragstr-params-path " + dragstr_model else ""} \
       -GQB 10 -GQB 20 -GQB 30 -GQB 40 -GQB 50 -GQB 60 -GQB 70 -GQB 80 -GQB 90 \
       ~{true="-ERC GVCF" false="" make_gvcf} \
       ~{bamout_arg}
@@ -128,12 +135,13 @@ task HaplotypeCaller_GATK4_VCF {
   >>>
 
   runtime {
-    docker: gatk_docker
+    docker: gatk_docker_to_use
     preemptible: preemptible_tries
-    memory: "6.5 GiB"
+    memory: "10 GiB"
     cpu: "2"
     bootDiskSizeGb: 15
     disks: "local-disk " + disk_size + " HDD"
+    maxRetries: 3
   }
 
   output {
--- a/tasks/broad/VariantCalling.wdl
+++ b/tasks/broad/VariantCalling.wdl
@@ -4,10 +4,12 @@ import "../../tasks/broad/GermlineVariantDiscovery.wdl" as Calling
 import "../../tasks/broad/Qc.wdl" as QC
 import "../../tasks/broad/Utilities.wdl" as Utils
 import "../../tasks/broad/BamProcessing.wdl" as BamProcessing
+import "../../tasks/broad/Dragen.wdl" as Dragen
 
 workflow VariantCalling {
 
   input {
+    Boolean run_dragen_mode = false
     File calling_interval_list
     File evaluation_interval_list
     Int haplotype_scatter_count
@@ -30,7 +32,26 @@ workflow VariantCalling {
 
   parameter_meta {
     make_bamout: "For CNNScoreVariants to run with a 2D model, a bamout must be created by HaplotypeCaller. The bamout is a bam containing information on how HaplotypeCaller remapped reads while it was calling variants. See https://gatkforums.broadinstitute.org/gatk/discussion/5484/howto-generate-a-bamout-file-showing-how-haplotypecaller-has-remapped-sequence-reads for more details."
+    run_dragen_mode: "Run variant calling using the DRAGEN-GATK pipeline, false by default."
   }
+
+  if (run_dragen_mode) {
+    call Dragen.ComposeSTRTableFile as ComposeSTRTableFile {
+       input:
+         ref_fasta = ref_fasta,
+         ref_fasta_idx = ref_fasta_index,
+         ref_dict = ref_dict,
+    }
+    call Dragen.CalibrateDragstrModel as DragstrAutoCalibration {
+       input:
+         ref_fasta = ref_fasta,
+         ref_fasta_idx = ref_fasta_index,
+         ref_dict = ref_dict,
+         alignment = input_bam,
+         str_table_file = ComposeSTRTableFile.str_table_file
+    }
+  }
+
 
   # Break the calling interval_list into sub-intervals
   # Perform variant calling on the sub-intervals, and then gather the results
@@ -67,7 +88,6 @@ workflow VariantCalling {
     }
 
     if (!use_gatk3_haplotype_caller) {
-
       # Generate GVCF by interval
       call Calling.HaplotypeCaller_GATK4_VCF as HaplotypeCallerGATK4 {
         input:
@@ -82,6 +102,8 @@ workflow VariantCalling {
           hc_scatter = hc_divisor,
           make_gvcf = make_gvcf,
           make_bamout = make_bamout,
+          run_dragen_mode = run_dragen_mode,
+          dragstr_model = DragstrAutoCalibration.dragstr_model,
           preemptible_tries = agg_preemptible_tries
        }
 
@@ -171,11 +193,11 @@ task MergeBamouts {
 
   Int disk_size = ceil(size(bams, "GiB") * 2) + 10
 
-  command {
+  command <<<
     samtools merge ~{output_base_name}.bam ~{sep=" " bams}
     samtools index ~{output_base_name}.bam
     mv ~{output_base_name}.bam.bai ~{output_base_name}.bai
-  }
+  >>>
 
   output {
     File output_bam = "~{output_base_name}.bam"
--- a/beta-pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
+++ b/beta-pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
@@ -38,13 +38,16 @@ import "../../../../../../structs/dna_seq/DNASeqStructs.wdl"
 # WORKFLOW DEFINITION
 workflow WholeGenomeGermlineSingleSample {
 
-  String pipeline_version = "2.3.1"
+  String pipeline_version = "3.0.0"
 
   input {
     SampleAndUnmappedBams sample_and_unmapped_bams
     DNASeqSingleSampleReferences references
+    DragmapReference? dragmap_reference
     VariantCallingScatterSettings scatter_settings
     PapiSettings papi_settings
+
+    File? dragmap_binary
 
     File? fingerprint_genotypes_file
     File? fingerprint_genotypes_index
@@ -53,6 +56,9 @@ workflow WholeGenomeGermlineSingleSample {
 
     Boolean provide_bam_output = false
     Boolean use_gatk3_haplotype_caller = true
+    Boolean run_dragen_mode = false
+    Boolean perform_bqsr = true
+    Boolean use_bwa_mem = true
   }
 
   # Not overridable:
@@ -67,6 +73,8 @@ workflow WholeGenomeGermlineSingleSample {
     input:
       sample_and_unmapped_bams    = sample_and_unmapped_bams,
       references                  = references,
+      dragmap_reference           = dragmap_reference,
+      dragmap_binary              = dragmap_binary,
       papi_settings               = papi_settings,
 
       contamination_sites_ud = references.contamination_sites_ud,
@@ -76,7 +84,9 @@ workflow WholeGenomeGermlineSingleSample {
       cross_check_fingerprints_by = cross_check_fingerprints_by,
       haplotype_database_file     = references.haplotype_database_file,
       lod_threshold               = lod_threshold,
-      recalibrated_bam_basename   = recalibrated_bam_basename
+      recalibrated_bam_basename   = recalibrated_bam_basename,
+      perform_bqsr                = perform_bqsr,
+      use_bwa_mem                 = use_bwa_mem
   }
 
   call AggregatedQC.AggregatedBamQC {
@@ -133,6 +143,7 @@ workflow WholeGenomeGermlineSingleSample {
 
   call ToGvcf.VariantCalling as BamToGvcf {
     input:
+      run_dragen_mode = run_dragen_mode,
       calling_interval_list = references.calling_interval_list,
       evaluation_interval_list = references.evaluation_interval_list,
       haplotype_scatter_count = scatter_settings.haplotype_scatter_count,
--- a/structs/dna_seq/DNASeqStructs.wdl
+++ b/structs/dna_seq/DNASeqStructs.wdl
@@ -20,6 +20,12 @@ struct ReferenceFasta {
   File ref_pac
 }
 
+struct DragmapReference {
+  File reference_bin
+  File reference_index_bin
+  File hash_table_cmp
+}
+
 struct DNASeqSingleSampleReferences {
   File contamination_sites_ud
   File contamination_sites_bed
--- a/tasks/broad/Qc.wdl
+++ b/tasks/broad/Qc.wdl
@@ -381,14 +381,14 @@ task ValidateSamFile {
     Array[String]? ignore
     Boolean? is_outlier_data
     Int preemptible_tries
-    Int memory_multiplier = 1
+    Int memory_multiplier = 2
     Int additional_disk = 20
   }
 
   Float ref_size = size(ref_fasta, "GiB") + size(ref_fasta_index, "GiB") + size(ref_dict, "GiB")
   Int disk_size = ceil(size(input_bam, "GiB") + ref_size) + additional_disk
 
-  Int memory_size = ceil(7 * memory_multiplier)
+  Int memory_size = ceil(10 * memory_multiplier)
   Int java_memory_size = (memory_size - 1) * 1000
 
   command {
@@ -406,6 +406,7 @@ task ValidateSamFile {
   runtime {
     docker: "us.gcr.io/broad-gotc-prod/picard-cloud:2.23.8"
     preemptible: preemptible_tries
+    maxRetries: 3
     memory: "~{memory_size} GiB"
     disks: "local-disk " + disk_size + " HDD"
   }
@@ -491,6 +492,7 @@ task CollectRawWgsMetrics {
     preemptible: preemptible_tries
     memory: "~{memory_size} GiB"
     disks: "local-disk " + disk_size + " HDD"
+    maxReplies: 3
   }
   output {
     File metrics = "~{metrics_filename}"
--- a/tasks/broad/SplitLargeReadGroup.wdl
+++ b/tasks/broad/SplitLargeReadGroup.wdl
@@ -16,6 +16,7 @@ version 1.0
 ## licensing information pertaining to the included programs.
 
 import "../../tasks/broad/Alignment.wdl" as Alignment
+import "../../tasks/broad/DragmapAlignment.wdl" as DragmapAlignment
 import "../../tasks/broad/BamProcessing.wdl" as Processing
 import "../../tasks/broad/Utilities.wdl" as Utils
 import "../../structs/dna_seq/DNASeqStructs.wdl" as Structs
@@ -32,11 +33,15 @@ workflow SplitLargeReadGroup {
     # (https://github.com/lh3/bwa/tree/master/bwakit),
     # listing the reference contigs that are "alternative".
     ReferenceFasta reference_fasta
+    DragmapReference? dragmap_reference
+
+    File? dragmap_binary
 
     Int compression_level
     Int preemptible_tries
     Int reads_per_file = 48000000
     Boolean hard_clip_reads = false
+    Boolean use_bwa_mem = true
   }
 
   call Alignment.SamSplitter as SamSplitter {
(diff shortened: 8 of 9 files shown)


michaelgatzen:
TODO change default docker image for GATK4 in GermlineVariantDiscovery to an image hosted on GCR

kbergin:
This is exciting!! Is it ok if I set up a time for you to walk through this PR with Lantern so we can get better acquainted? George will add you to a list so that the tests will run on your PR.

michaelgatzen:
retest this please

tasks/broad/Dragen.wdl
```diff
@@ -0,0 +1,102 @@
...
+  String base_name = basename(alignment)
+  String out_file_name = base_name + ".dragstr"
+  Int disk_size_gb = ceil(size([ref_fasta, ref_fasta_idx, alignment, str_table_file], "GiB")) + 
+                        20 # 20 for the rest of the fs.
+
+  String parallel_args  = if (threads <= 1) then "" else "--parallel" 
+
+  command <<<
+    set -x
+    samtools index -@ ~{threads} ~{alignment}
```
ldgauthier, line 79: It's not obvious to me why we can't pass in the index instead.  The indexing step was slow as cold molasses when I ran it on Terra recently.

tasks/broad/Dragen.wdl
```diff
@@ -0,0 +1,102 @@
...
+    gatk --java-options "-Xmx2g -DGATK_STACKTRACE_ON_USER_EXCEPTION=true" \
```
ldgauthier, line 80: I couldn't get this to run on a cram without also passing -Dsamjdk.reference_fasta=~{ref_fasta} in the java options, so until we can get that resolved in GATK, we might want to add that since this task claims to handle bam or cram.

tasks/broad/Dragen.wdl
michaelgatzen: You're absolutely right, I don't know either why it wasn't originally just passed as an input, thanks for your input!

tasks/broad/Dragen.wdl
michaelgatzen: Great, thanks for finding that! In the pipeline this task is only ever run with BAMs as input from a previous step, but I don't see any harm in passing that argument.

tasks/broad/VariantCalling.wdl
```diff
@@ -30,8 +32,28 @@ workflow VariantCalling {
...
+    run_dragen_mode: "Run variant calling using the DRAGEN-GATK pipeline, false by default."
   }
 
+  if (run_dragen_mode) {
+    call Dragen.ComposeSTRTableFile as ComposeSTRTableFile {
+       input:
+         ref_fasta = ref_fasta,
+         ref_fasta_idx = ref_fasta_index,
+         ref_dict = ref_dict,
+    }
```
michaelgatzen, line 44: Calculate this once per reference and pass it as input

jessicaway:
Ok to test

michaelgatzen:
retest this please

jessicaway:
retest this please

jessicaway:
More tests to make sure I haven't broken the testing on forks:
retest this please

michaelgatzen:
It looks like there are some (expected?) differences between the eval and truth files which probably originate from the switch to a more recent GATK version. I will look into this in depth, but is there a process in place to validate these differences and consequently update the truth accordingly?

kbergin:
@ldgauthier is probably best to comment on her process. In the past we've provided a spreadsheet of the differences in metrics observed between the two versions as a starting point. I believe the team has a script for pulling those metrics into a sheet - @gbggrant or @jessicaway can comment perhaps. It may be that all of the validation that the methods has been doing is sufficient to approve the change.

We do have a way for updating the truth data when it's been approved. George or Jess or someone else from Lantern can also help there!

kbergin:
Is the test comparing the DRAGEN path through the pipeline to the old version or is it comparing the same path and there are some slight differences? 

I ask because I saw there are some parameters along the lines of 'use dragen' in the code that seem to push data through the dragen path, and so I imagine there needs to be a corresponding update to the test inputs to use that path? Or we should create new test inputs/a new test for that path specifically?

michaelgatzen:
> Is the test comparing the DRAGEN path through the pipeline to the old version or is it comparing the same path and there are some slight differences?

That's a good question, the tests use the exact same code path as before since all the DRAGEN-GATK-related arguments are disabled by default. So I suspect the differences to originate from switching from GATK 4.1.8.0 to 4.2.0.0.
That being said, the tests currently only really evaluate exome, since by default the WGS pipeline still uses GATK3, so the tests don't cover the GATK4 changes (and therefore succeed).

Going forward, we may want to decide how to run tests with different code paths. Currently there are 4 dimensions that alter the path: BWA/DragMap, BQSR on/off, GATK3/GATK4, Dragen mode on/off. So if the tests should cover all paths, we would need 16 (minus 4, because we can't run GATK3 in Dragen mode) cases, which seems like a lot of work. Or do people have alternative ideas?

kbergin:
Excellent ok, thanks for clarifying. In that case, using the script I mentioned (from @jessicaway and @gbggrant) to pull together the metrics showing the differences observed in exomes between GATK 4.1.8.0 and GATK 4.2.0.0 to share with @ldgauthier would be the next step. Overall, this is definitely the first test needed to confirm that current production releases of the pipelines only change how we expect them to so that the PR at least can be unblocked and we can figure out DRAGEN related tests later.

I imagine you're seeing pipeline release validation test failures because the code that you've changed touches more than the genome pipeline. We have that test to make sure that the pipeline version is bumped for any shared tasks that are changed. They probably aren't major version changes like whole genomes. The team can help you identify the pipeline versions that need to be bumped if you need! 

Speaking of version updates, do we think this would be a major version change for WGS now while the default path through the code is still the same? Or maybe later when we deprecate the previous default paths? A question for the group! We've been using this definition of major/minor version [here](https://broadinstitute.github.io/warp/documentation/About_WARP/VersionAndReleasePipelines.html#versioning-requirements).

Regarding testing multiple paths, that's a great topic for discussion. I hope those I've tagged in this comment will put their thoughts here.

ldgauthier:
Can we make sure to check the scatter count?  I'm running this now on a typical WGS cram and the HC task is taking more than 27 hours: https://job-manager.dsde-prod.broadinstitute.org/jobs/134074c0-fc0c-4595-909e-f49f2e35515b

michaelgatzen:
@ldgauthier Does this pipeline with the DRAGEN-GATK changes behave differently than without? Because it shouldn't, the scatters can be set as a workflow argument and nothing in this PR should touch that code... Would be good to know if it does!

ldgauthier:
I'm doubling the scatters in another run now.  The slow one was $3.30 (allegedly), which isn't too bad, but it isn't great. It looks like about half the shards were able to run preemptible, but I think with half the runtime (double scatter count) we should be able to do better.  I'll run one with the vanilla 4.2 pipeline for comparison.

tasks/broad/Dragen.wdl
```diff
@@ -0,0 +1,102 @@
...
+  String base_name = basename(alignment)
+  String out_file_name = base_name + ".dragstr"
+  Int disk_size_gb = ceil(size([ref_fasta, ref_fasta_idx, alignment, str_table_file], "GiB")) + 
+                        20 # 20 for the rest of the fs.
+
+  String parallel_args  = if (threads <= 1) then "" else "--parallel" 
+
+  command <<<
+    set -x
+    gatk --java-options "-Xmx2g -DGATK_STACKTRACE_ON_USER_EXCEPTION=true -Dsamjdk.reference_fasta=~{ref_fasta}" \
```
ldgauthier, line 80: This isn't enough in a lot of cases.  Mem should be exposed as a param.

tasks/broad/Dragen.wdl
```diff
@@ -0,0 +1,102 @@
...
+      CalibrateDragstrModel \
+        -R ~{ref_fasta} \
+        -I ~{alignment} \
+        -str ~{str_table_file} \
+        -O ~{out_file_name} \
+        ~{parallel_args} \
+        --verbosity DEBUG
```
ldgauthier, line 87: We should take this out before we merge -- the stderr is insanely long.

tasks/broad/Dragen.wdl
michaelgatzen: Right, it looks like (almost) none of the JVM memory limits are exposed in WARP, but I can certainly do that here. Thanks for noticing. From your experience, do you have a point of reference on what would work as a default value?

tasks/broad/Dragen.wdl
michaelgatzen: Fixed, thanks. Note also that CalibrateDragstrModel is still  marked as a beta tool. Should we change that @vruano?

vruano: I don't know what the criterion is to decide when a beta label should be dropped. @droazen is probably the one we should be asking that.

tasks/broad/Dragen.wdl
vruano: Doing some single sample testing its seems that parallel does not work well in the input is a cram file unless the memory is push up, perhaps as much as 2gb for each intended parallel thread. And even so it might be nearly as fast to transform the cram into bam (using. samtools view -@ thread-count...). bam with or without parallelism does not seem to be affected and in that particular case they can run with the default 2gb without issues.

michaelgatzen: Thanks for your input. It seems like in this pipeline this task is only ever called with BAM input, so I wonder why Laura would run into issues. For me, this task has also never failed, but if tasks fail we should absolutely fix it. @ldgauthier could you maybe confirm if this task failed for you within this pipeline or when executed separately (e.g. on a CRAM)?

vruano: if accepts cram we should make it does work with them, that reasonable assumption is what got @ldgauthier in trouble.

vruano: I have posted in my fork changes for the calibrate task to fix the issue with cram: https://github.com/vruano/warp/pull/1

This is based on some testing I performed on a. problematic sample data HG00544

michaelgatzen: Great, thanks @vruano! I added the changes to this branch. I also gave you access to this repo, feel free to merge or PR-merge into this branch as well

ldgauthier: It turns out the OOM issue is cram-related: https://github.com/broadinstitute/gatk/issues/7189#issuecomment-817806008
My suggestion is we restrict input to bams (just in the docs, nothing fancy)

ldgauthier: But do still bubble up the mem argument

ldgauthier:
I've been running this workflow on a bunch of the 1kg samples and the cost is about where I expect (~$2.50), but shards 7, 9, and 10 are taking a loooooong time (on the order of 10h, compared with ~2hr for most of the rest).  Strangely they're not getting preempted, but it's something to keep an eye on @kbergin

beta-pipelines/broad/somatic/single_sample/targeted/TargetedSomaticSingleSample.wdl
```diff
@@ -173,7 +173,7 @@ workflow TargetedSomaticSingleSample {
     File? fingerprint_detail_metrics = AggregatedBamQC.fingerprint_detail_metrics
 
     File duplicate_metrics = UnmappedBamToAlignedBam.duplicate_metrics
-    File output_bqsr_reports = UnmappedBamToAlignedBam.output_bqsr_reports
+    File? output_bqsr_reports = UnmappedBamToAlignedBam.output_bqsr_reports
```
jamesemery, line 176: test

matthdsm:
Question,

I'm not sure this PR is still active, but could you explain some of the choices you made in `DragmapAlignment`?
- Why use `samtools fastq` instead of `picard SamToFastq`?
- Why avoid the use of pipes?
- Why convert to fastq altogether, since `dragmap` accepts bam as input?

Thanks for the clarifications. We're benchmarking a workflow of our own and I'd be interested in hearing the reasons behind these choices.

Cheers
M

michaelgatzen:
@matthdsm

Thank you so much for your input, Matthias! You raise interesting points, that was in fact an oversight from my side. Beta versions of Dragmap only supported FASTQ input and we had asked Illumina to implement BAM input, so I must have missed that they did that in their most recent version.

As for the piping, the way that Dragmap's input/output buffers are set up they do not support piping (or at least they didn't for previous versions). I will test it again with this version and if it works we can pipe directly from Dragmap to MergeBamAlignments.

Again, thank you for your valuable input, this is very helpful!

matthdsm:
Hi Michael,

It strikes me as a bit odd that bam input wasn't support from the get go. Wasn't that one of the selling points of the aligner in the presentations of the Dragen-GATK workflow?
Anyways, thank you for the work on this. Looking forward to hearing if everything works as expected.

Also, was there a reason for using samtools instead of picard for the bam to fastq conversion? performance or resource wise?

Cheers
M

tasks/broad/Dragen.wdl
```diff
@@ -0,0 +1,113 @@
...
+## LICENSING :
+## This script is released under the WDL source code license (BSD-3) (see LICENSE in
+## https://github.com/broadinstitute/wdl). Note however that the programs it calls may
+## be subject to different licenses. Users are responsible for checking that they are
+## authorized to run all programs before running this script. Please see the docker
+## page at https://hub.docker.com/r/broadinstitute/genomes-in-the-cloud/ for detailed
+## licensing information pertaining to the included programs.
+
+
+task ComposeSTRTableFile {
```
jessicaway, line 20: This appears to be based solely on the reference files (rather than unique input files). Is this something we should do every time we run the pipeline, or could we consider generating this file as a reference file that is passed in to the pipeline as an input?

tasks/broad/GermlineVariantDiscovery.wdl
```diff
@@ -92,8 +92,15 @@ task HaplotypeCaller_GATK4_VCF {
     Boolean make_bamout
     Int preemptible_tries
     Int hc_scatter
-    String gatk_docker = "us.gcr.io/broad-gatk/gatk:4.1.8.0"
+    Boolean run_dragen_mode = false
```
jessicaway, line 95: Same comment as before about defaulting once unless there is a reason to do otherwise

tasks/broad/GermlineVariantDiscovery.wdl
```diff
@@ -209,6 +219,43 @@ task HardFilterVcf {
   }
 }
 
+# This hard filtering matches DRAGEN 3.4.12. For later DRAGEN versions, this needs to be updated.
+task DragenHardFilterVcf {
```
jessicaway, line 223: Should this be included in the DragenTasks wdl? (I'm fine with whatever others think is most appropriate, just wanted to raise the question)

tasks/broad/GermlineVariantDiscovery.wdl
```diff
@@ -209,6 +219,43 @@ task HardFilterVcf {
...
+      -O ~{output_vcf_name}
+  }
+  output {
+    File output_vcf = "~{output_vcf_name}"
+    File output_vcf_index = "~{output_vcf_name}.tbi"
+  }
+  runtime {
+    docker: gatk_docker
+    preemptible: preemptible_tries
+    memory: "3 GiB"
```
jessicaway, line 253: Another place to consider including memory in the inputs or scaling based on input size. If this is not a very memory hungry task then hard coding a small value may be fine

tasks/broad/GermlineVariantDiscovery.wdl
```diff
@@ -130,7 +140,7 @@ task HaplotypeCaller_GATK4_VCF {
   runtime {
     docker: gatk_docker
     preemptible: preemptible_tries
-    memory: "6.5 GiB"
+    memory: "~{memory_size} GiB"
```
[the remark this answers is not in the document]
jessicaway: You are absolutely right that changing the memory in the task inputs will break call caching for the task. We generally don't worry about this issue since you would rarely manually change the memory on a task that was passing (and therefore could benefit from call caching). It is important to note though!

As far as a variable affecting Cromwell's memory retry feature, I don't know the answer to this one. I don't see any indication in the documentation that this would be a problem, but I have reached out to the batch team to confirm. I will update here when I have an answer

jessicaway (commented):
@michaelgatzen This looks fantastic! Thanks for all your work on this. My comments are mostly minor, but feel free to reach out if you have any questions

jessicaway:
@ldgauthier @michaelgatzen 
I wanted to point out that we are planning to add reclocking as the default for WGS and exomes (see [Laura's open PR](https://github.com/broadinstitute/warp/pull/141)). Should reblocking be added to the workflow when run in Dragen mode as well?

tasks/broad/GermlineVariantDiscovery.wdl
```diff
@@ -209,6 +219,43 @@ task HardFilterVcf {
...
 
+# This hard filtering matches DRAGEN 3.4.12. For later DRAGEN versions, this needs to be updated.
+task DragenHardFilterVcf {
+  input {
+    File input_vcf
+    File input_vcf_index
+    Boolean make_gvcf
+    String vcf_basename
+    Int preemptible_tries
+    String gatk_docker = "us.gcr.io/broad-gatk/gatk:4.1.8.0"
```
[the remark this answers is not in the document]
jessicaway: Might as well, this will need to be validated anyways

tasks/broad/GermlineVariantDiscovery.wdl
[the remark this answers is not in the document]
jessicaway: Batch has confirmed that this will **not** affect the memory retry feature in cromwell

tasks/broad/Alignment.wdl
```diff
@@ -32,6 +32,7 @@ task SamToFastqAndBwaMemAndMba {
     Int compression_level
     Int preemptible_tries
     Boolean hard_clip_reads = false
+    Boolean unmap_contaminant_reads = true
```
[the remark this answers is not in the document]
michaelgatzen: If set to true this does not change any default behavior. People can set it to false for maximum concordance with hardware DRAGEN, but it is a sensible step and we recommend leaving it on

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
```diff
@@ -38,11 +38,12 @@ import "../../../../../../structs/dna_seq/DNASeqStructs.wdl"
 # WORKFLOW DEFINITION
 workflow WholeGenomeGermlineSingleSample {
 
-  String pipeline_version = "2.3.5"
+  String pipeline_version = "2.3.6"
```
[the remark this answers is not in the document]
michaelgatzen: This is something that Lantern should decide

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
```diff
@@ -53,6 +54,12 @@ workflow WholeGenomeGermlineSingleSample {
 
     Boolean provide_bam_output = false
     Boolean use_gatk3_haplotype_caller = true
+    Boolean run_dragen_mode = false
```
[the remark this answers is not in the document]
michaelgatzen: Yes (even though you can't use dragen mode with GATK3 obviously). In order to be functionally equivalent, users must set use_gatk3_haplotype_caller to false, run_dragen_mode to true, use_bwa_mem to false, use_spanning_event_genoytping to false and perform_bqsr to false. They might want to enable/disable features individually though if they're not interested in functional equivalence. It might make sense to add an additional single argument that automatically overrides all these options if people care about functional equivalence. If you think that's a good idea feel free to add that!

tasks/broad/DragmapAlignment.wdl
```diff
@@ -0,0 +1,98 @@
...
+      PROGRAM_GROUP_NAME="dragen-os" \
+      UNMAPPED_READ_STRATEGY=COPY_TO_TAG \
+      ALIGNER_PROPER_PAIR_FLAGS=true \
+      UNMAP_CONTAMINANT_READS=~{unmap_contaminant_reads} \
+      ADD_PG_TAG_TO_READS=false
+  >>>
+  runtime {
+    docker: "us.gcr.io/broad-dsde-methods/dragmap:1.2.0"
+    preemptible: preemptible_tries
+    memory: "40 GiB"
```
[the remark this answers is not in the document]
michaelgatzen: Moved to input section. Illumina recommends a fixed memory of 40 GB for Dragmap

tasks/broad/Dragen.wdl
michaelgatzen: This is done once per reference, and we're planning on uploading it to the public references bucket. Should we include this task in the WDL then or not?

tasks/broad/GermlineVariantDiscovery.wdl
michaelgatzen: Fine either way, up for Lantern to decide I would say

tasks/broad/GermlineVariantDiscovery.wdl
michaelgatzen: I used the same configuration as the HardFilterVcf task right above this task. Let me know if you want me to change it.

tasks/broad/Alignment.wdl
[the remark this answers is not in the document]
michaelgatzen: Also I had it set at every level for backwards compatibility, if e.g. users and/or other workflows in this repo would call a subworkflow WDL, this would break because the new option hasn't been set. I changed all the added options now to be only set at the WholeGenomeGermlineSingleSample WDL, let me know if I should change it back @jessicaway

pipelines/broad/dna_seq/germline/variant_calling/VariantCalling.wdl
```diff
@@ -4,12 +4,15 @@ import "../../../../../tasks/broad/GermlineVariantDiscovery.wdl" as Calling
...
 import "../../../../../tasks/broad/Utilities.wdl" as Utils
 import "../../../../../tasks/broad/BamProcessing.wdl" as BamProcessing
+import "../../../../../tasks/broad/Dragen.wdl" as Dragen
 
 workflow VariantCalling {
 
   String pipeline_version = "1.0.1"
 
   input {
+    Boolean run_dragen_mode = false
```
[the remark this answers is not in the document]
michaelgatzen: Same reasoning here, I set it for backwards compatibility if people call the VariantCalling.wdl directly

tasks/broad/GermlineVariantDiscovery.wdl
michaelgatzen: Same question about backwards compatibility here

tasks/broad/Alignment.wdl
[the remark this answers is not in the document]
michaelgatzen: Actually, that is exactly what happened as the womtool validation for ExomeReprocessing and ExternalExomeReprocessing are failing - perform_bqsr (also an added option that I had previously set a default value on all subworkflow levels) wasn't defined as an input in ExomeGermlineSingleSample.wdl

michaelgatzen:
We need to hold on merging this PR, as we found a bug in DRAGMAP that either needs to get fixed or a workaround needs to be implemented.

tasks/broad/Alignment.wdl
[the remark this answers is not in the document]
jessicaway: sometimes it makes sense to thread it all the way through (in which case you will have a default for every top level pipeline, or you have a default at each level that a pipeline uses directly, which would allow backwards compatibility without changing the top level pipeline). If you think it will rarely be changed, or only changed by very well-informed users, you can set it at the task level input with a default. In this case you could still change it from the inputs json, but you have to know that it is an option to change.

I think either decision is reasonable and I can help look at this closer if you like.

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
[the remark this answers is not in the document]
jessicaway: @kbergin do you have a preference for the version increase here?

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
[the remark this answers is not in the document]
jessicaway: we might want to add a check to ensure you don't run dragen mode with GATK3, otherwise I think we can just document the different options available

tasks/broad/Dragen.wdl
jessicaway: If the files are available in the public bucket, I think it would make more sense to include them as input reference files rather than regenerate them each time. If we think that users may want to use non-standard references we could either provide documentation for how to generate these files or we could include the input files as optional and only run this task if the input is not provided

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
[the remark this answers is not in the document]
kbergin: I'd say minor version increase for this work for now, though I think we're going to increase the major version soon anyways with the reblocking outputs change.

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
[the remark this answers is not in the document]
kbergin: @michaelgatzen could you include with this PR an example inputs json that will run the pipeline in DRAGEN mode with all the correct parameters set?

tasks/broad/Dragen.wdl
kbergin: Agreed, it should not be in the WDL if it is generated once. You can pull the code to generate it out into a separate script or documentation though, if you think we'll want to regenerate it occasionally.

tasks/broad/Alignment.wdl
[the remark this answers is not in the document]
michaelgatzen: @jessicaway Yes, I think this case is somewhat special, because on one side we expect this option to be changed frequently (e.g. whether to use BWA mem or Dragmap is a pretty fundamental difference), so we would certainly want to be able to set that at the top level WholeGenomeSingleSample WDL. On the other side, we don't want to break backwards compatibility for all the existing pipelines out there. So I think that defaulting it at each level (to the old value) makes the most sense in this case, but please let me know if you have a better option!

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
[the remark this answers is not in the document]
michaelgatzen: @kbergin The user education team is working on a comprehensive documentation for these options, but sure, I can provide an inputs json with the options. Where do you want me to put it, and which input data should it use?

tasks/broad/Dragen.wdl
michaelgatzen: Ok, I removed the task from the WDL. The STR table file is already passed as an (optional) input file, so no other changes are required.

tasks/broad/GermlineVariantDiscovery.wdl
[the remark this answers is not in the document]
michaelgatzen: So this doesn't need to be changed and I can resolve this comment, correct?

tasks/broad/DragmapAlignment.wdl
```diff
@@ -0,0 +1,102 @@
...
+## https://github.com/broadinstitute/wdl). Note however that the programs it calls may
+## be subject to different licenses. Users are responsible for checking that they are
+## authorized to run all programs before running this script. Please see the docker
+## page at https://hub.docker.com/r/broadinstitute/genomes-in-the-cloud/ for detailed
+## licensing information pertaining to the included programs.
+
+import "../../structs/dna_seq/DNASeqStructs.wdl"
+
+# Read unmapped BAM, convert on-the-fly to FASTQ and stream to BWA MEM for alignment, then stream to MergeBamAlignment
+task SamToFastqAndDragmapAndMba {
```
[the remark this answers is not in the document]
michaelgatzen: I'm happy to change this, but I would like to explain my reasoning behind having DragmapAlignment.wdl separate from the other DragenTasks.wdl. The DragmapAlignment.wdl is very similar to the Alignment.wdl (which performs BWA mem alignment), so depending on what aligner you choose it either calls Alignment.wdl or DragmapAlignment.wdl. Of course I can move the SamToFastqAndDragmapAndMba task to DragenTasks.wdl, I guess it just wouldn't be as analogous to Alignment.wdl that way. Happy to do whatever you prefer!

tasks/broad/GermlineVariantDiscovery.wdl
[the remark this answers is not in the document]
jessicaway: correct!

tasks/broad/Alignment.wdl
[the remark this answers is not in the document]
jessicaway: That's fine with me

michaelgatzen:
Illumina published their most recent Dragmap patch (version 1.2.1) which resolved the previously mentioned issue, so this is not a blocker anymore.

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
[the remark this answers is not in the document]
kbergin: Let's make a dragen fe and maximum performance version of this example inputs. It can stay in the same folder 
https://github.com/broadinstitute/warp/blob/develop/pipelines/broad/dna_seq/germline/single_sample/wgs/input_files/WholeGenomeGermlineSingleSample.inputs.plumbing.json

tasks/broad/DragmapAlignment.wdl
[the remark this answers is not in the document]
jessicaway: I'm fine either way. 
@ldgauthier which do you prefer?

tasks/broad/GermlineVariantDiscovery.wdl
jessicaway: I would say, if we decide to move the alignment task to DragenTasks and keep all the dragen code together, let's move this one too. If not, it makes sense for the dragen version to live next to the non-dragen equivalent

tasks/broad/Dragen.wdl
kbergin: What's the latest on this thread?

kbergin:
As an update, I chatted with David Roazen and @michaelgatzen today about reblocking. We've decided reblocking should be on by default for DRAGEN in both run modes as that is our best practice recommendation. Depending on PR merge order, let's make sure the right PR covers the DRAGEN mode reblocking outputs by default. cc @ldgauthier

pipelines/broad/dna_seq/germline/single_sample/wgs/WholeGenomeGermlineSingleSample.wdl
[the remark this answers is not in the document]
michaelgatzen: Done, but not resolving until the reference files have been moved to Google's bucket and we verify that the example inputs are working

tasks/broad/Dragen.wdl
michaelgatzen: The memory argument has been exposed, so if I understand correctly then we can resolve this thread? Or are there concerns about this task as it stands right now @ldgauthier ? For reference, this task now lives in https://github.com/michaelgatzen/warp/blob/3e41edc997460dfd42dc53b4cbef4731c0751eda/tasks/broad/DragenTasks.wdl#L19

jessicaway:
Apologies for the testing comments, I am trying to get the scientific tests to run

jessicaway:
retest this please

ldgauthier:
> @ldgauthier @michaelgatzen I wanted to point out that we are planning to add reclocking as the default for WGS and exomes (see [Laura's open PR](https://github.com/broadinstitute/warp/pull/141)). Should reblocking be added to the workflow when run in Dragen mode as well?

Good point.  I will try to remember that when I rebase my PR.

tasks/broad/DragmapAlignment.wdl
```diff
@@ -0,0 +1,100 @@
...
+      PROGRAM_GROUP_VERSION="${DRAGMAP_VERSION}" \
+      PROGRAM_GROUP_COMMAND_LINE="dragen-os -b ~{input_bam} -r dragen_reference --interleaved=1" \
+      PROGRAM_GROUP_NAME="dragen-os" \
+      UNMAPPED_READ_STRATEGY=COPY_TO_TAG \
+      ALIGNER_PROPER_PAIR_FLAGS=true \
+      UNMAP_CONTAMINANT_READS=~{unmap_contaminant_reads} \
+      ADD_PG_TAG_TO_READS=false
+  >>>
+  runtime {
+    docker: "us.gcr.io/broad-dsde-methods/dragmap:1.2.1"
```
ldgauthier, line 90: This needs to move out of dsde-methods

jessicaway:
The following pipelines need versions and changelogs updated:
`ExomeGermlineSingleSample`
`WholeGenomeGermlineSingleSample`
`VariantCalling`
`ExomeReprocessing`
`WholeGenomeReprocessing`
`ExternalExomeReprocessing`
`ExternalWholeGenomeReprocessing`

Let's make sure we incorporate Laura's wording in the changelogs for the Exome pipelines ("In accordance with known improvements in GATK 4.1.9.0 and 4.2.0.0, sensitivity to phased variants is improved in a small number of cases and genotypes are more accurate in a very small number of cases involving indels and spanning deletions.")

@ekiernan @michaelgatzen can you help coordinate the changelog notes for this? The PR description included here is REALLY helpful, but it might be a little long to copy and paste into all the changelogs 

I am happy to have a meeting next week to figure this out if that is helpful

pipelines/broad/dna_seq/germline/single_sample/exome/ExomeGermlineSingleSample.wdl
```diff
@@ -129,6 +129,7 @@ workflow ExomeGermlineSingleSample {
       ref_fasta = references.reference_fasta.ref_fasta,
       ref_fasta_index = references.reference_fasta.ref_fasta_index,
       ref_dict = references.reference_fasta.ref_dict,
+      ref_str = references.reference_fasta.ref_str,
```
jessicaway, line 132: Not needed

pipelines/broad/dna_seq/germline/single_sample/exome/ExomeGermlineSingleSample.changelog.md
```diff
@@ -1,3 +1,8 @@
+# 2.6.0
+2021-10-18
+* Updated GATK to v4.2.2.0 for variant calling
```
jessicaway, line 3: Should we incorporate Laura's explanation of the observed changes that this update produces here? 
```suggestion
* Updated GATK to v4.2.2.0 for variant calling
    - In accordance with known improvements in GATK 4.1.9.0 and 4.2.0.0, sensitivity to phased variants is improved in a small number of cases and genotypes are more accurate in a very small number of cases involving indels and spanning deletions.
```

jessicaway (approved):

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
 
-Learn more about the software tools implemented in these tasks by reading the GATK [data pre-processing](https://gatk.broadinstitute.org/hc/en-us/articles/360035535912) and [germline short variant discovery](https://gatk.broadinstitute.org/hc/en-us/articles/360035535932) documentation.
 
+| Input name | Input description | Input type |
+| --- | --- | --- |
+| fingerprint_genotypes_file | Genotype VCF if optionally performing fingerprinting.  | File |
+| fingerprint_genotypes_index | Optional index for the fingerprinting VCF. | File |
+| wgs_coverage_interval_list | Interval list for the CollectWgsMetrics tool. | File |
+| provide_bam_output | If set to true, provides the aligned BAM and index as workflow output; default set to false. | Boolean |
+| use_gatk3_haplotype_caller | Uses the GATK3.5 HaplotypeCalller; default set to true. | Boolean |
```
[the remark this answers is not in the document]
jessicaway: @michaelgatzen is that right? This has to be false to run in Dragen mode, correct?

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
+| dragen_functional_equivalence_mode | Boolean used to run the WGS pipeline in a mode functionally equivalent to DRAGEN; set to false by default. | Boolean |
+| dragen_maximum_quality_mode | Boolean used to run the pipeline in DRAGEN mode with modifications to produce maximum quality results; set to false by default. | Boolean |
```
jessicaway, line 128: What happens if I set `dragen_maximum_quality_mode` to `true` AND `dragen_functional_equivalence_mode` to `true`?

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
+| run_dragen_mode_variant_calling | Boolean used to indicate that DRAGEN mode should be used for variant calling; default set to false but must be true to compose DRAGstr model and perform variant calling with HaplotypeCaller in dragen-mode. | Boolean |
+| use_spanning_event_genotyping | Boolean used to call the HaplotypeCaller --disable-spanning-event-genotyping parameter; default set to true so that variant calling includes spanning events. Set to false to run the DRAGEN pipeline.  | Boolean |
+| unmap_contaminant_reads | Boolean to indicate whether to identify extremely short alignments (with clipping on both sides) as cross-species contamination and unmap the reads; default set to true. This feature is not used in the pipeline mode functionally equivalent to DRAGEN. | Boolean |
+| perform_bqsr | Boolean to turn on base recalibration with BQSR; default set to true, but not necessary when running the pipeline in DRAGEN mode. | Boolean |
+| use_bwa_mem | Boolean indicating if workflow should use the BWA mem aligner; default set to true, but must be set to false to alternatively run the DRAGEN-GATK DRAGMAP aligner. | Boolean |
```
jessicaway, line 133: Do we have documentation that lists all the valid combinations of inputs and why you might want to configure it different ways? 

It's late in the life of this PR to bring this up, but if we are validating and supporting only 3 modes of the pipeline for instance, would it make sense to abstract these inputs into a single input, for instance `pipeline_mode` that takes 1 of 3 options (for instance ` GATK_best_prcatices`, `DRAGEN_functional_equivalence`, or `DRAGEN_maximu_quality`) and then these variables are all set by the workflow to the appropriate values for that mode?

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
+| unmap_contaminant_reads | Boolean to indicate whether to identify extremely short alignments (with clipping on both sides) as cross-species contamination and unmap the reads; default set to true. This feature is not used in the pipeline mode functionally equivalent to DRAGEN. | Boolean |
+| perform_bqsr | Boolean to turn on base recalibration with BQSR; default set to true, but not necessary when running the pipeline in DRAGEN mode. | Boolean |
+| use_bwa_mem | Boolean indicating if workflow should use the BWA mem aligner; default set to true, but must be set to false to alternatively run the DRAGEN-GATK DRAGMAP aligner. | Boolean | 
+| use_dragen_hard_filtering | Boolean that indicates if workflow should perform hard filtering using the GATK VariantFiltration tool with the --filter-name "DRAGENHardQUAL"; default set to false. | Boolean
+|  read_length | Set to a max of 250 for collecting WGS metrics; specified in the workflow WDL, not in the input JSON. | Int |
```
jessicaway, line 135: What does it mean it is specified in the workflow, not the inputs? Essentially this is hardcoded and would require a WDL edit to change?

jessicaway: Same question for "inputs" below.

If these are not actually inputs, perhaps we should make another section to describe them. It makes sense to call out assumed values, but if they are not configurable than it might make sense to group them separately

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
+ 
+## Workflow outputs
+
+The table below describes the final workflow outputs. If running the workflow on Cromwell, these outputs are found in the respective task's execution directory.
+
+
+| Output variable name | Description | Type |
+| --- | --- | --- |
+| quality_yield_metrics | The quality metrics calculated for the unmapped BAM files. | Array of files |
+| unsorted_read_group_base_distribution_by_cycle_pdf | PDF of the base distribution for each unsorted, readgroup-specific BAM. | Array of PDFs |
```
jessicaway, line 263: This is probably a dumb question, but is an array of pdfs different than an array of files? Did we just want to note the specific file type here? (not suggesting this needs to change, just curious)

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
+| unsorted_read_group_insert_size_histogram_pdf | Histograms of insert size for the unsorted, readgroup-specific BAMs.  | Array of PDFs |
+| unsorted_read_group_insert_size_metrics | Insert size metrics for the unsorted, readgroup-specific BAMs. | Array of files |
+| unsorted_read_group_quality_by_cycle_pdf | Quality by cycle PDF for the unsorted, readgroup-specific BAMs. | Array of PDFs |
+| unsorted_read_group_quality_by_cycle_metrics | Quality by cycle metrics for the unsorted, readgroup-specific BAMs.  | Array of Files |
+| unsorted_read_group_quality_distribution_pdf | Quality distribution PDF for the unsorted, readgroup-specific BAMs.  | Array of PDFs |
+| unsorted_read_group_quality_distribution_metrics | Quality distribution metrics for the unsorted, readgroup-specific BAMs. | Array of PDFs |
+| read_group_alignment_summary_metrics | Alignment summary metrics for the aggregated BAM. | File |
+| read_group_gc_bias_detail_metrics | GC bias detail metrics for the aggregated BAM. | File |
+| read_group_gc_bias_pd | PDF of the GC bias by readgroup for the aggregated BAM. | PDF |
```
jessicaway, line 273:
```suggestion
| read_group_gc_bias_pdf | PDF of the GC bias by readgroup for the aggregated BAM. | PDF |
```

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
+| agg_gc_bias_pdf | PDF of GC bias for the aggregated BAM. | PDF |
+| agg_gc_bias_summary_metrics | GC bias summary metrics for the aggregated BAM. | File |
+| agg_insert_size_histogram_pdf | Histogram of insert size for aggregated BAM. | PDF |
+| agg_insert_size_metrics | Insert size metrics for the aggregated BAM. | File |
+| agg_pre_adapter_detail_metrics | Details metrics for artifacts that occur prior to the addition of adaptors for the aggregated BAM. | File |
+| agg_pre_adapter_summary_metrics | Summary metrics for artifacts that occur prior to the addition of adaptors for the aggregated BAM. | File |
+| agg_quality_distribution_pdf | PDF of the quality distribution for the aggregated BAM. | PDF |
+| agg_quality_distribution_metrics | Quality distribution metrics for the aggregated BAM. | File |
+| agg_error_summary_metrics | Error summary metrics for the aggregated BAM. | File | 
+| fingerprint_summary_metrics | Optional fingerprint summary metrics for the aggregated BAM.. | File |
```
jessicaway, line 292:
```suggestion
| fingerprint_summary_metrics | Optional fingerprint summary metrics for the aggregated BAM. | File |
```

website/docs/Pipelines/Whole_Genome_Germline_Single_Sample_Pipeline/README.md
```diff
@@ -44,57 +89,235 @@ The [Whole Genome Germline Single Sample workflow](https://github.com/broadinsti
...
+| fingerprint_detail_metrics |Optional fingerprint detail metrics for the aggregated BAM. | File |
```
jessicaway, line 293:
```suggestion
| fingerprint_detail_metrics | Optional fingerprint detail metrics for the aggregated BAM. | File |
```

PR #204, opened 2021-01-14
GL-1362 GDC somatic whole genome pipeline accepts alligned cram

jessicaway:
Added the option to accept an aligned cram or bam as input to the GDC somatic whole genome pipeline. 

The previous input (a single unmapped bam) is still valid


kbergin:
I really appreciate that we were able to update the pipeline to act appropriately based on the input passed in. The flexibility really saves us some future work I think! Thanks Jess!

jessicaway:
There are still a couple open questions:
1) Should we take multiple unmapped bams instead of a single unmapped bam as input? (our germline wgs and exome pipelines except multiple unmapped bams)
2) Do we want an input check to print some pretty error message in case you don't specify any input? (currently it will just fail when it tries to access an undefined file)

gbggrant (approved):

khajoue2 (approved):

kbergin:
@jessicaway I think leaving it with a single one is fine for now, I have no specific use case at this time for this pipeline accepting more than one input file, so it seems like something we could add if the need arises.

On question 2, do we do an input check in other pipelines like that? How fast would the pipeline fail if someone didn't put in an input?

jessicaway:
It would fail pretty quickly, but since each individual input is optional it would not fail instantly (as it would with a missing required input). 
It seems ok to me to leave it for now. I would guess users would be able to figure out what happened if they forgot to specify an input, even if the error message isn't specific

kbergin:
Ah I see, yea probably fine to leave this PR as it is I think!

PR #229, opened 2021-02-08
GL-1424 java memory for cram to unmapped bams

jessicaway:


pipelines/broad/reprocessing/cram_to_unmapped_bams/CramToUnmappedBams.wdl
```diff
@@ -104,8 +104,10 @@ task RevertSam {
     Int memory = 3
```
gbggrant, line 104: Maybe name this 'memory_in_GiB or something like that?  I could see someone passing 3000.
And perhaps we should standardize on mb?  (not sure if this is already an use like this)

gbggrant (approved):

pipelines/broad/reprocessing/cram_to_unmapped_bams/CramToUnmappedBams.wdl
jessicaway: I can rename to clarify. 
I see both mb and gb for memory specification in warp. I agree we should standardize. Which one is preferred?

gbggrant: I think mb, that we can stick with ints?

jessicaway: Sure, that sounds good. I will just confirm with Hornet and Hops that switching it here is not going to cause issues for them

nikellepetrillo (approved):

PR #230, opened 2021-02-08
GDC somatic pipeline - return contamination value as a file 

jessicaway:


jessicaway:
The tests do not automatically run for beta-pipelines. I have manually kicked off the test; it can be found [here](https://gotc-jenkins.dsp-techops.broadinstitute.org/job/warp-workflow-tests/4542/)

gbggrant (approved):

tbl3rd (approved):
Thank you.  Is `contam.txt` the traditional name?

jessicaway:
This code came from the tag group, it looks like they named it `contam.txt`. I can rename it if there is another name you prefer

tbl3rd:
> This code came from the tag group, it looks like they named it `contam.txt`. I can rename it if there is another name you prefer

Works for me.

PR #248, opened 2021-02-23
GL-1432 Optimus will not fail when empty drops has problem running

khajoue2:
Emptydrops tool fails if the sample has few cells ( we expect a couple of thousand cells). We decided to let the pipeline finish the processing for such samples and instead produce NAs in empty_drop_results.csv file and finally in the loom file.


ekiernan:
Documentation updates for this PR are [here](https://github.com/broadinstitute/warp/pull/251)

dockers/skylab/emptydrops/emptyDropsWrapper/emptyDropsWrapper.R
```diff
@@ -198,9 +198,29 @@ tryCatch({
...
+        emptyDrops_result <- matrix(data=NA,nrow=n_rows,ncol=7)
+        ## Convert output from DataFrame to data.frame
+        emptyDrops_result <- as.data.frame(emptyDrops_result)
+        colnames(emptyDrops_result) <-  c("CellId","Total", "LogProb", "PValue", "Limited", "FDR", "IsCell")
+        emptyDrops_result[,"CellId"] = colnames(inputMatrix)
+        ## Write the output file
+        catv('Writing output CSV...')
+        write.csv(x=emptyDrops_result, file=output_csv,row.names=FALSE)
+        catv('done\n')
+        quit(save="no",status=17)
```
jessicaway, line 216: You will probably need to catch this error in the command block in wdl, since a non-zero return code is usually interpreted as an error

khajoue2: I changed it to status 0 so the wdl can transfer the output NA file from the vm to the next task.

kishorikonwar (approved):
Looks good to me.

jessicaway (approved):

PR #268, opened 2021-03-18
Promote VariantCalling to be a top-level pipeline

tlangs:
(h)Ops has a need for a VariantCalling workflow that can be run independently of alignment. This PR promotes the VariantCalling portion of the pipeline to be a top-level workflow that can be released.


kbergin:
Thanks so much for this Trevyn! I'll make sure the team reviews Monday (most folks are out today)

tlangs:
I think there's some work to be done in the smart tests to actually run the VariantCalling tests that I coded up.

tests/broad/scala_test/src/main/scala/org/broadinstitute/dsp/pipelines/commandline/ConfigParser.scala
```diff
@@ -810,4 +810,22 @@ class ConfigParser
           )
         }
     )
+
+  note("")
+  germlineCloudPipelineCommandLineConfig(
+    VariantCalling, { config =>
+      (config.test, config.germlineCloudConfig.category) match {
+        case (VariantCalling, WorkflowTestCategory.Scientific) =>
+          failure(
```
gbggrant, line 819: Should it just run the plumbing instead then?  (and not fail).

gbggrant (approved):
Looks good.

tests/broad/scala_test/src/main/scala/org/broadinstitute/dsp/pipelines/commandline/ConfigParser.scala
tlangs: I think that's dishonest to the user. If scientific tests are requested, I think the test suite should just say "Sorry Dave, I can't do that." and fail. I worry that making "scientific" tests run plumbing data might bite us in the future if we forget that its actually plumbing data and rely on it for something.

gbggrant: Or maybe we should just run the scientific tests here then?  And not just say they're covered by another test.

tlangs: That would require extracting the intermediate BAM files from all the Scientific WGS and Exome samples and creating input files. It can be done, but I didn't really have time to do so before transitioning teams.

gbggrant: Fair enough.  I'll bring it up with the team / make a ticket.

kbergin:
Correct me if I'm wrong, but I think the last step for this PR is review by @ldgauthier ? 

And a rebase it looks like.

gbggrant:
I think we need another developer review too.

jessicaway (approved):

jessicaway:
can we add this to the smart test by adding the arguments to this file? 
warp/scripts/get_changed_pipeline_worklow_test_args.sh

ldgauthier (approved):
This seems straightforward.  I'm fine with not having scientific tests specifically for variant calling since it's covered by single-sample.

jessicaway:
retest this please

PR #272, opened 2021-03-25
Updating Germline Genome Single Sample Pipeline plumbing data to use broad-public-datasets

kbergin:
Trying out changing the plumbing test data for genomes to the equivalent files in broad public datasets.

Tests should confirm if they are in fact the same samples given their name is the same.


pshapiro4broad:
add to whitelist

jessicaway (approved):

nikellepetrillo (approved):

PR #292, opened 2021-05-04
Gl 1474 sn-ss2 build indices

khajoue2:
Added docker file and scripts to modify gtfs and add introns for snSS2 pipeline


kishorikonwar (approved):
Looks good to me.

nikellepetrillo (approved):

pipelines/skylab/build_indices/BuildIndices.wdl
```diff
@@ -322,7 +374,7 @@ workflow BuildIndices {
 
   # version of this pipeline
 
-  String pipeline_version = "0.0.1"
+  String pipeline_version = "0.1.0"
```
kbergin, line 377: Usually pipeline version is at the top of the WDL, is it down on this line because we declare the tasks first?

dockers/skylab/snss2-build-indices/add-introns-to-gtf.py
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env python3
```
kbergin, line 1: Is a .py a docker? should this be in the scripts folder?

dockers/skylab/snss2-build-indices/modify_gtf_human.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
```
kbergin, line 1: Same Q as above

dockers/skylab/snss2-build-indices/modify_gtf_mouse.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
kbergin, line 1: One more time

pipelines/skylab/build_indices/BuildIndices.wdl
khajoue2: Yes, it should be inside the workflow and we did not have any imports in this pipeline so the version is down where the workflow block is defined.

dockers/skylab/snss2-build-indices/add-introns-to-gtf.py
khajoue2: All dockers in skylab have their scripts next to then if they have scripts in the docker. It makes it easier to change the docker in case we want to modify those scripts.

kbergin: Ok, thanks

pipelines/skylab/build_indices/BuildIndices.wdl
kbergin: Ok!

kbergin (approved):

PR #293, opened 2021-05-05
Gl 1435 optimus star solo

kishorikonwar:
STAR solo based Optimus pipeline


kbergin:
Do you plan to merge this into the main Optimus pipeline? I'm surprised to see it separate as a Optimus-starsolo pipeline. I do not want us to duplicate the pipeline like that.

PR #294, opened 2021-05-06
Dev -> Staging

jessicaway:


nikellepetrillo (approved):

jessicaway:
The changelog and release validation are failing for GCD somatic WGS. The changes included are not sufficient to warrant a version increase (added a new scientific input, which we do not consider part of the pipeline and therefore don't version, and added a single blank line the the WLD file)

PR #309, opened 2021-05-27
Remove PAIRED_RUN parameter from MergeBamAlignment

wjdingman1:


kbergin:
@wjdingman1 would you add a small description to the PR including linking the user issue in WARP.

I think your release validation is going to fail because there will be required updates to pipeline version numbers and changelogs to complete releasing this PR. The team can help you look at the failure log, interpret it, and make the required changes.

wjdingman1:
@kbergin so even though this task is not specific to any single pipeline, the pipelines using it will still need their changelogs updated?

kbergin:
This task is used in multiple pipelines, even though this parameter isn't, so I think the change to the command would need to be reflected as a change to the pipelines. It's just a patch change to the version number. @jessicaway lmk if I'm misrepresenting this process.

kbergin:
@kachulis now I'm curious - what did this parameter used to do and does the Picard tool just do it by default now?

jessicaway:
add to whitelist

kachulis:
@kbergin I assume it used to be needed when running with paired end data, probably because of how the tool used to be written.  It has been deprecated for a LONG time, and wasn't doing anything even before it was deprecated (back to at least 2013).  Now the tool just doesn't need to be told whether the data is paired ended or single ended, and can handle the data fine on its own either way.

kachulis (approved):

jessicaway:
My previous comment is a specific phrase for Jenkins. 

The version and changelog validation has now run. The pipelines affected by the change can be found in the console output for the pipeline release validation job, in this case:
`14:59:23 ExomeGermlineSingleSample.wdl has not been changed and needs updating
14:59:23 WholeGenomeGermlineSingleSample.wdl has not been changed and needs updating
14:59:23 ExomeReprocessing.wdl has not been changed and needs updating
14:59:23 ExternalExomeReprocessing.wdl has not been changed and needs updating
14:59:23 ExternalWholeGenomeReprocessing.wdl has not been changed and needs updating
14:59:23 WholeGenomeReprocessing.wdl has not been changed and needs updating`

The version for each of these will need to be changed. I agree with @kbergin that this will be a patch. Each changelog will also need a new entry. The changelog style guide can be found [here](https://github.com/broadinstitute/warp/blob/develop/changelog_style.md)

wjdingman1:
@jessicaway Ah okay great, can I get access to view those Jenkins logs at some point? 

Will make all these changelog updates first thing next week!

gbggrant (approved):

jessicaway (approved):

PR #333, opened 2021-06-21
[GL-835] Duplicate docker image in QC pipeline

cheyenne-gold:
This pr addresses [this issue](https://github.com/broadinstitute/dsde-pipelines/issues/734) on the dsde-pipelines repo. Since we've since moved to the warp repo, it's being fixed here. The CheckPreValidation task has two defined docker images, this pr removes the duplicate.


cheyenne-gold:
This PR needs to be from/to develop, not trivy-action.

PR #337, opened 2021-06-24
[GL-204] Docker Images best practices

wjdingman1:
This PR ensures that best practices are being followed for our broad images (excluding GoTC at the moment)
- Images are alpine based (VerifyBamID has to use debian base because of specific packages needed)
- Single run steps or multi stage builds are used
- All images are running tini as their default entry point - please read [here](https://github.com/krallin/tini/issues/8) for explanation
- Removing cached package index files
- Steps in dockerfile are properly documented
- Images are labeled with the proper maintainer
- Image tags are semantically versioned and include a timestamp (moving back to unix time)


jessicaway (approved):
Looks good, Thanks!

khajoue2 (approved):
LGTM.

PR #340, opened 2021-06-25
[GL-1575] Samtools specific docker images

wjdingman1:
This PR creates a samtools specific docker image available at `us.gcr.io/broad-gotc-prod/samtools` and replaces tasks _(Utilities.wdl & CramToUnmappedBam.wdl)_ that were previously using the GoTC docker image while only needing samtools.


jessicaway (approved):
Looks good!

cheyenne-gold (approved):

PR #353, opened 2021-07-13
GL-1621.  Make chip_well_barcode and analysis_version_number available as outputs

gbggrant:
Make chip_well_barcode and analysis_version_number available as outputs


gbggrant:
Not wild about this approach at all.

jessicaway (approved):
Seems Ok. Why do we need this?

gbggrant:
Hornet would like these values as part of the workflow outputs.  I'm not really sure why they can't just parse the metadata (which has the inputs in it).  Does seem kind of too much.

nikellepetrillo (approved):

PR #362, opened 2021-07-19
[GL-1488] updating input_name for SmartSeq2LoomOutput to be in quotes

cheyenne-gold:
Attempting to  fix create_loom_ss2.py failing to parse input_name correctly when is a space is in the input_name by wrapping the variable in quotes.


jessicaway (approved):

nikellepetrillo (approved):
looks good!

PR #388, opened 2021-08-11
Updated GDC Readme

ekiernan:
-Added description of output file location
-Updated syntax to match Google style guide


website/docs/Pipelines/Genomic_Data_Commons_Whole_Genome_Somatic/README.md
```diff
@@ -38,79 +38,82 @@ For the latest workflow version and release notes, please see the [changelog](ht
...
 | bam_readgroup_to_contents | view | [Samtools](http://www.htslib.org) | Extracts all the readgroups from the BAM header and returns a WDL array where each row is a readgroup. |
-| biobambam_bamtofastq | bamtofastq | [biobambam2](https://gitlab.com/german.tischler/biobambam2) | Converts the uBAMs to FASTQ |
-| emit_pe_records/emit_se_records | --- | --- | Associates the fasta file(s) generated by biobambam_bamtofastq with their respective readgroup; creates an array of structs to be used as input for alignment |
-| bwa_pe/ bwa_se | BWA mem, view | [BWA](http://bio-bwa.sourceforge.net/) and [Samtools](http://www.htslib.org) | Aligns FASTQ reads to the reference genome, generating an aligned BAM file |
+| biobambam_bamtofastq. | bamtofastq | [biobambam2](https://gitlab.com/german.tischler/biobambam2) | Converts the uBAMs to FASTQ |
```
ekiernan, line 84:
```suggestion
| biobambam_bamtofastq | bamtofastq | [biobambam2](https://gitlab.com/german.tischler/biobambam2) | Converts the uBAMs to FASTQ. |
```

website/docs/Pipelines/Genomic_Data_Commons_Whole_Genome_Somatic/README.md
```diff
@@ -38,79 +38,82 @@ For the latest workflow version and release notes, please see the [changelog](ht
...
+| collect_insert_size_metrics | [CollectInsertSizeMetrics](https://gatk.broadinstitute.org/hc/en-us/articles/360037055772) | [Picard](https://broadinstitute.github.io/picard/) | Generates metrics about insert size distribution in the form of both a histogram and table (txt file). |
 
-## Workflow Outputs
+## Workflow outputs
+The following table describes the workflow outputs. If running the workflow using Cromwell, these outputs will automatically be placed in the respective task execution directory.
 
-| Output Name | Description | Type |
+Alternatively, Cromwell allows you to specify an output directory using an options.json, as described in Cromwell's [Workflow Options Overview](https://cromwell.readthedocs.io/en/stable/wf_options/Overview/)(see the section on [Output Copying](https://cromwell.readthedocs.io/en/stable/wf_options/Overview/#output-copying)).
```
kbergin, line 97: So this information applies to every pipeline documented on the WARP doc site. It seems like a lot to copy the content in every single doc.. and a bit odd to only put it in this one (unless it's already in all the others?). Do we have a quickstart for running pipelines in Cromwell outside of Terra? If we do, it seems like this would be better to put there?

ekiernan: Yes, I made a ticket to add it to each pipeline as I didn't think the lift would be that big. We don't currently have a quick start, but that sounds like a good idea too.

kbergin (approved):

jessicaway (approved):

PR #394, opened 2021-08-16
Wreleaser changelog

wjdingman1:
Just adding a CHANGELOG.md to the wreleaser tool so I can cut a release later on.

### DIFF
--- a/wreleaser/CHANGELOG.MD
+++ b/wreleaser/CHANGELOG.MD
@@ -1,0 +1,4 @@
+# 0.1.0
+2021-07-29 (Date of Last Commit)
+
+* Initial Release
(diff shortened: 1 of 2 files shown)


ekiernan (approved):
LGTM! Curious about how we decide if a tool should start v1.0.0 or 0.1.0?

wjdingman1:
> LGTM! Curious about how we decide if a tool should start v1.0.0 or 0.1.0?

Its usually if a tool is being used in 'production' and has a stable API... guess for this case it doesn't really matter since its  an accessory tool and not exactly critical

There are a couple more features I would like to add then will probably cut a v1.0.0 release

cheyenne-gold (approved):

PR #420, opened 2021-09-01
remove scientific test option for GDC somatic pipeline

nikellepetrillo:
remove scientific test option for GDC somatic pipeline, also remove scientific test input json


nikellepetrillo:
closing because we now have scientific data

PR #421, opened 2021-09-07
Remove some comments and dead code

wjdingman1:


projects/hca_mvp/tasks/AdapterTasks.wdl
```diff
@@ -509,6 +509,7 @@ task GetReferenceDetails {
   }
 }
 
+# This task is specifically for getting intermediate level loom ids for Optimues
```
cheyenne-gold, line 512:
```suggestion
# This task is specifically for getting intermediate level loom ids for Optimus
```

projects/hca_mvp/tasks/AdapterTasks.wdl
```diff
@@ -574,6 +575,8 @@ task CopyToStagingBucket {
   }
 }
 
+# This task is specifically for getting the MergeLoomsOptimus version
```
cheyenne-gold, line 578:
```suggestion
# This task is specifically for getting the MergeOptimusLooms version
```

projects/hca_mvp/tasks/AdapterTasks.wdl
```diff
@@ -574,6 +575,8 @@ task CopyToStagingBucket {
...
+# MergeLoomsOptimus version is found in MergeOptimusLooms.wdl
```
cheyenne-gold, line 579:
```suggestion
# MergeOptimusLooms version is found in MergeOptimusLooms.wdl
```

projects/smartseq2/CreateSs2AdapterMetadata.wdl
```diff
@@ -105,14 +105,8 @@ workflow CreateSs2AdapterMetadata {
   String multi_sample_pipeline_version = ParseSs2CromwellMetadata.pipeline_version
   String single_sample_pipeline_version = ParseSs2CromwellMetadata.single_sample_pipeline_version
 
-  call Tasks.GetSs2PipelineVersion as CheckPipelineVersion {
```
cheyenne-gold, line 108: I have a vague memory of @nikellepetrillo mentioning we run some tests just as fail safe tasks that will fail if something is off. But I also might be thinking of project_name... Before we delete this I just want to make sure it's not necessary anymore?

nikellepetrillo: yup it was project_name, i think this can be removed!

nikellepetrillo (approved):

projects/smartseq2/CreateSs2AdapterMetadata.wdl
wjdingman1: yeah i think we added this just copying off of Optimus....Optimus needs this task because we actually need to append the `1.0.0` pipeline_version from MergeLooms to the `v_`

Since we are grabbing both the versions from the metadata.json in ss2 then we can get rid of it

PR #432, opened 2021-09-14
Create build script for HCA preprocessing and update tag to 2.0

wjdingman1:


pipelines/skylab/optimus/scripts/docker_versions.tsv
```diff
@@ -0,0 +1 @@
+quay.io/humancellatlas/hca_preprocessing:1.0
```
cheyenne-gold, line 1: Has this script not been run yet? I guess I was expecting to see the 2.0 version in this tsv as well

cheyenne-gold (approved):

pipelines/skylab/optimus/scripts/docker_build.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
jessicaway, line 1: should this live with the other docker build scripts (i.e. in the dockers folder instead of the pipelines folder)?

jessicaway (approved):

pipelines/skylab/optimus/scripts/docker_build.sh
wjdingman1: probably should go in the dockers/skylab/ folder but we haven't gotten around to refactoring all that just yet

PR #441, opened 2021-09-22
GL-1725. Have Arrays Pipeline lookup Bead Pool Manifest File by Name

gbggrant:


tasks/broad/Utilities.wdl
```diff
@@ -206,3 +206,18 @@ task SumFloats {
     preemptible: preemptible_tries
   }
 }
+
```
jessicaway, line 209: This is going to conflict with the Imputation pipeline :( oh well

jessicaway (approved):

tasks/broad/Utilities.wdl
gbggrant: Nope - should be the exact same method (I lifted it from Imputation's PR)

nikellepetrillo (approved):
lgtm!

PR #443, opened 2021-09-23
Three new GDA samples for Arrays testing.

gbggrant:
Three new GDA samples for Arrays testing.
These are technically control samples (An Ashkenazim trio), but we don't have control data set up
for these, so they can't be run as controls.


nikellepetrillo (approved):

jessicaway (approved):
This is probably a silly question... but the descriptions says 3 new samples, but it seems there are 6 new files. Is that correct?

cheyenne-gold (approved):

gbggrant:
@jessicaway yes, three new samples, 6 files (3 for IlluminaGenotypingArray scientific tests and 3 for Arrays scientific tests).  Same data though.

PR #449, opened 2021-09-27
GL-1736. Don't write control fingerprints to the Mercury Fingerprint Store

gbggrant:
GL-1736. Don't write fingerprints to the Mercury Fingerprint Store if the sample is a control


jessicaway (approved):
good idea!

cheyenne-gold (approved):

PR #451, opened 2021-09-28
GL-1730. Have Arrays Pipeline look up Extended Illumina Manifest

gbggrant:
GL-1730. Have Arrays Pipeline (optionally) look up Extended Illumina Manifest from lookup table.

### DIFF
--- a/tasks/broad/InternalArraysTasks.wdl
+++ b/tasks/broad/InternalArraysTasks.wdl
@@ -366,3 +366,72 @@ task GetNextArraysQcAnalysisVersionNumber {
     Int analysis_version_number = read_int(stdout())
   }
 }
+
+task ResolveExtendedIlluminaManifestFile {
+  input {
+    File extended_manifest_map_file
+    String bpm_filename
+    String egt_filename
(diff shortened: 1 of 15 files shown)


tasks/broad/InternalArraysTasks.wdl
```diff
@@ -366,3 +366,36 @@ task GetNextArraysQcAnalysisVersionNumber {
     Int analysis_version_number = read_int(stdout())
   }
 }
+
+task ResolveExtendedIlluminaManifestFile {
+  input {
+    File extended_manifest_map_file
+    String bpm_filename
+    String egt_filename
```
jessicaway, line 374: Is the cluster file being used other than the error messages?

gbggrant: Not yet.  It probably should be used to help discriminate - good catch.  I was on the fence about it, but I think it's better to include it in the grep than not.

jessicaway (approved):
Looks good 👍

cheyenne-gold (approved):

PR #454, opened 2021-10-01
Some cleanups / Improvements in Arrays

gbggrant:
Some Cleanups.
1) Don't pass down the autocall_version - rather return it from autocall (this is in there for the special case where autocall fails and doesn't return its version, but we still wanted to track it).
2) Break UploadEmptyArrayMetrics into its own method so that there are less exposed task level optional inputs.
3) Modify vault token handling. Localize token to the VM and then read it directly into an environment variable to prevent the token ever being viewable in the metadata, etc.

### DIFF
--- a/tasks/broad/IlluminaGenotypingArrayTasks.wdl
+++ b/tasks/broad/IlluminaGenotypingArrayTasks.wdl
@@ -522,6 +522,13 @@ task AutoCall {
 
   String gtc_filename = "~{chip_well_barcode}.gtc"
 
+  # This is the autocall_version, It is normally output by autocall (gencall) itself, except for the case
+  # where autocall fails (likely due to normalization errors)
+  # In this case it no longer emits the version in its output, so we have it here so that it can be output and
+  # stored in the database.
+  # NB - this should be returned from the docker ideally.
+  String autocall_ver = "3.0.0"
+
   command <<<
     set -e
     rm -rf ~{chip_well_barcode}
@@ -550,6 +557,7 @@ task AutoCall {
 
   output {
     File gtc_file = gtc_filename
+    String autocall_version = autocall_ver
   }
 }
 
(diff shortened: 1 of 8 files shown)


tasks/broad/IlluminaGenotypingArrayTasks.wdl
```diff
@@ -522,6 +522,13 @@ task AutoCall {
 
   String gtc_filename = "~{chip_well_barcode}.gtc"
 
+  # This is the autocall_version, It is normally output by autocall (gencall) itself, except for the case
+  # where autocall fails (likely due to normalization errors)
+  # In this case it no longer emits the version in its output, so we have it here so that it can be output and
+  # stored in the database.
+  # NB - this should be returned from the docker ideally.
+  String autocall_ver = "3.0.0"
```
jessicaway, line 530: Is there a way to read this from the software in the docker instead of hard-coding it?

jessicaway (approved):
Loos good! Just 1 minor comment

tasks/broad/IlluminaGenotypingArrayTasks.wdl
gbggrant: Unfortunately, it doesn't look like we can.  If autocall succeeds and generates a GTC file, you can query it for the autocallVersion `gtcFile.getAutoCallVersion()` - that's where we get the '3.0.0' for passing runs.  But if autocall fails, it doesn't produce a gtc.  If you query IAAP (the program that runs on the docker and is running autocall) - you get a different version number:
`root@786be8f0813f:/usr/gitc# /usr/gitc/iaap/iaap-cli/iaap-cli -v
1.1.0-1.1.0-sha.80d7e5b3d9c1fdfc2e99b472a90652fd3848bbc7`

jessicaway: fair enough

cheyenne-gold (approved):

PR #461, opened 2021-10-08
Np standardize runtime imputation tasks

nikellepetrillo:


tasks/broad/ImputationTasks.wdl
```diff
@@ -5,17 +5,18 @@ task CalculateChromosomeLength {
     File ref_dict
     Int chrom
     String ubuntu_docker
```
cheyenne-gold, line 7: I believe we discussed defining the default docker definition here and removing the block in imputation.wdl
```suggestion
    String ubuntu_docker = "ubuntu:20.04"
```

pipelines/broad/arrays/imputation/Imputation.wdl
```diff
@@ -74,8 +66,7 @@ workflow Imputation {
         input_vcfs = select_first([single_sample_vcfs]),
         input_vcf_indices = select_first([single_sample_vcf_indices]),
         output_vcf_basename = "merged_input_samples",
-        bcftools_docker = bcftools_docker_tag,
-        mem = merge_ssvcf_mem_gb
+        memory_mb = merge_ssvcf_mem_gb
```
cheyenne-gold, line 69: we are trying to use memory in mb but merge_ssvcf_mem_gb is set to 3. i'm not sure how the task handles it, but should merge_ssvcf_mem_gb be increased here?

tasks/broad/ImputationTasks.wdl
```diff
@@ -4,18 +4,19 @@ task CalculateChromosomeLength {
...
-  Int disk_size = ceil(2*size(ref_dict, "GiB")) + 5
-
   command {
     grep -P "SN:~{chrom}\t" ~{ref_dict} | sed 's/.*LN://' | sed 's/\t.*//'
   }
   runtime {
     docker: ubuntu_docker
-    disks: "local-disk " + disk_size + " HDD"
-    memory: "2 GiB"
+    disks: "local-disk " + disk_size_gb + " HDD"
```
cheyenne-gold, line 17: sometimes disk is being set (in other pipelines) using `disks: "local-disk ${disk_size_gb} HDD"` syntax. is one better than the other? do we care to make it all consistent? just curious!

tasks/broad/ImputationTasks.wdl
```diff
@@ -307,25 +331,26 @@ task OptionalQCSites {
     File input_vcf
     File input_vcf_index
     String output_vcf_basename
-    String bcftools_vcftools_docker
+    String bcftools_vcftools_docker = "us.gcr.io/broad-dsde-methods/imputation_bcftools_vcftools_docker:v1.0.0"
     Float? optional_qc_max_missing
     Float? optional_qc_hwe
+    Int cpu = 1
```
cheyenne-gold, line 337: in other places it looks like the docker variables (docker, cpu, memory and disk) are below the other task inputs and separated by a blank line. is that a style we want to implement while cleaning this up? I personally like the look, but probably just a preference not really important

nikellepetrillo: i like the look of that as well, it helps to separate runtime attribute inputs from the others. will update 👍

pipelines/broad/arrays/imputation/Imputation.wdl
nikellepetrillo: going to update this so merge_ssvcf_mem_gb is changed to merge_ssvcf_mem_mb in the actual wdl, and i will set that to 3000 mb instead of 3 gb, just to keep it all consistent!

pipelines/broad/arrays/imputation/Imputation.changelog.md
```diff
@@ -1,3 +1,8 @@
+#1.0.1
```
wjdingman1, line 1: we should probably just update the date of last commit for 1.0.0 since imputation hasn't been released to prod yet for 1.0.0

nikellepetrillo: good call, will update

PR #467, opened 2021-10-14
GL-1678. Install the variant_rsids_file in the google-hosted public bucket

gbggrant:
Install the variant_rsids_file in the google-hosted public bucket.


kbergin:
Wow we have so many inputs jsons for this pipeline !

kbergin (approved):

PR #485, opened 2021-10-28
Azure specific changes for exome and wgs pipeline

olesya13:
1. Made azure specific changes to runtime attributes:
a.  _disks: "local-disk " + disk_size + " HDD"_ converted to  _disk: disk_size + " GB"_
b.  _preemptible: preemptible_tries_  converted to _preemptible: true    maxRetries:  preemptible_tries_. With TES the _preemptible_ attribute is a boolean (not an integer). You can specify preemptible as true or false for each task, when set to true Cromwell on Azure will use a low-priority batch VM to run the task; consider adding _maxRetries_ to keep the retry functionality
c.  _bootDiskSizeGb: "15"_ removed, since attribute _bootDiskSizeGb_ is not supported by the TES backend. Attribute _zones_ is not supported too.
2. Removed scatter coefficient in disk_size calculation in HaplotypeCaller, BaseRecalibrator and ApplyBQSR (we expect that this change will be reverted when NIO is available).

Runtime attribute changes are needed due to different Cromwell backends, you can find more info about them here https://cromwell.readthedocs.io/en/develop/backends/TES/ and https://cromwell.readthedocs.io/en/develop/RuntimeAttributes/

With changes described above exome and wgs pipelines ran on Azure for earlier version (before Dragen components were added),  Similar changes will be required in DragenTasks.wdl and DragmapAlignment.wdl

### DIFF
--- a/pipelines/broad/dna_seq/germline/variant_calling/VariantCalling.wdl
+++ b/pipelines/broad/dna_seq/germline/variant_calling/VariantCalling.wdl
@@ -221,8 +221,9 @@ task MergeBamouts {
   runtime {
     docker: "biocontainers/samtools:1.3.1"
     memory: "4 GiB"
-    disks: "local-disk ~{disk_size} HDD"
-    preemptible: 3
+    disk: "~{disk_size} GB"
+    preemptible: true
+    maxRetries:  3
     cpu: 1
   }
 }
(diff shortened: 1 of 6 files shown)


khajoue2 (approved):
lgtm

kbergin:
@khajoue2 thanks for taking a look, this PR won't be merged right now, it's mostly for sharing the changes needed to run the pipelines on Azure. But as it is, it breaks the pipelines on GCP (thus why tests are failing). I've added a DO NOT MERGE label.

jessicaway:
@olesya13 Is this PR still needed for testing?

olesya13:
@jessicaway there is no need to merge it. This PR was created to demonstrate what changes needed to run pipelines on Azure. All of this is reviewed with Broad and documented. There is no need to keep it, I think.

PR #491, opened 2021-11-02
Updated Optimus Overview

ekiernan:
1. Changed the link to Optimus release
2. Included description of pseudogene handling vs. Cell Ranger
3. Added validation report

### DIFF
--- a/website/docs/Pipelines/Optimus_Pipeline/README.md
+++ b/website/docs/Pipelines/Optimus_Pipeline/README.md
@@ -6,7 +6,7 @@ sidebar_position: 1
 
 | Pipeline Version | Date Updated | Documentation Author | Questions or Feedback |
 | :----: | :---: | :----: | :--------------: |
-| [optimus_v5.0.0](https://github.com/broadinstitute/warp/releases) | August, 2021 | [Elizabeth Kiernan](mailto:ekiernan@broadinstitute.org) | Please file GitHub issues in warp or contact [Kylee Degatano](mailto:kdegatano@broadinstitute.org) |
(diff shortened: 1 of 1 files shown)


khajoue2 (approved):
Looks good!

website/docs/Pipelines/Optimus_Pipeline/README.md
```diff
@@ -96,6 +96,8 @@ The JSON file also contains metadata for the reference information in the follow
 | use_strand_info | Optional string for reading stranded data. Default is "false"; set to "true" to count reads in stranded mode. | "true" or "false" (default) |
 | emptydrops_lower | UMI threshold for emptyDrops detection; default is 100. | NA |
 
+#### Pseudogene handling
+Optimus reference files are downloaded directly from GENCODE (see Quickstart table) and are not modified to remove pseudogenes. This is in contrast to the [references created for Cell Ranger](https://support.10xgenomics.com/single-cell-multiome-atac-gex/software/release-notes/references#header) which remove pseudogenes and small RNAs.
```
kbergin, line 100: I hope I don't say this wrong - but I think we also do something special with multimap reads to pseudogenes because we keep them in the reference, which also introduces some differences between the pipelines. @kishori82 or @khajoue2 can you clarify?

kbergin (changes requested):

website/docs/Pipelines/Optimus_Pipeline/README.md
khajoue2: The aligner maps reads to the entire genome and the bam files get annotated with the positions we specify in the gtf files. When we count reads, if the gtf includes psudogenes, a read can have more than one annotation maps, one to a coding gene and another to a psudogene. In cellranger, they don't include psudogenes so such reads are not considered as multi-mapped. We discard multimapped reads so for this case, we do not count it towards the final count value.

ekiernan: @khajoue2 This sounds like it should produce the same results then?

ekiernan: Ok, updated the wording to reflect the different results between the two pipelines.

kbergin (approved):

PR #501, opened 2021-11-09
Imputation TDR index workaround

jessicaway:
TDR does not store indices in the same "file path" as the file they correspond to. This creates issues for some of our tasks that infer the index path from the corresponding file path. 

This PR copies that index files next to the corresponding vcfs within the VM for files being pulled from TDR. 

We had hoped that TDR would have a fix for this problem on their end, but it seems that may not be coming anytime soon, so rather than run off a branch we suggest merging and releasing this change as part of the pipeline. 

The issue is discussed further in the following slack threads:
https://broadinstitute.slack.com/archives/C021CDSQ83X/p1633638988206600?thread_ts=1633636244.199200&cid=C021CDSQ83X
and
https://broadinstitute.slack.com/archives/CESEYJW9W/p1635434130125500


tasks/broad/ImputationTasks.wdl
```diff
@@ -80,10 +80,10 @@ task CountVariantsInChunks {
     File panel_vcf
     File panel_vcf_index
 
-    Int disk_size_gb = ceil(2*size([vcf, vcf_index, panel_vcf, panel_vcf_index], "GiB"))
     String gatk_docker = "us.gcr.io/broad-gatk/gatk:4.1.9.0"
     Int cpu = 1
     Int memory_mb = 4000
+    Int disk_size_gb = 2 * ceil(size([vcf, vcf_index, panel_vcf, panel_vcf_index], "GiB")) + 20
```
cheyenne-gold, line 86:
```suggestion
    Int disk = 2 * ceil(size([vcf, vcf_index, panel_vcf, panel_vcf_index], "GiB")) + 20
```
wondering if it's worth updating this to match the variable naming we've worked towards with standardizing our runtime blocks?

cheyenne-gold: and then updating line 98 to `disks: "local-disk ${disk} HDD"`

cheyenne-gold (approved):

tasks/broad/ImputationTasks.wdl
jessicaway: Hmmm, I see your point, but the naming is consistent throughout all the tasks in this file, so I am hesitant to change just this one. 
Also do we want to standardize onf including or excluding  `_gb`?

cheyenne-gold: Ah, I see what you mean. You're totally right, good to be consistent in the file. I believe we decided on excluding `_gb`. We captured the criteria from that mobbing discussion here: https://github.com/broadinstitute/warp/pull/464#issue-1025407499, I should probably track down a good README or documentation to add these guidelines to.

jessicaway: No worries, thanks for checking!!

wjdingman1 (approved):

pipelines/broad/arrays/imputation/Imputation.changelog.md
```diff
@@ -1,3 +1,9 @@
+# 1.0.1
+2021-11-09 (Date of Last Commit)
+* Added a step to MergeSingleSampleVcfs task to copy index files next to VCF files before running merge
```
gbggrant, line 3:
```suggestion
* Added a step to MergeSingleSampleVcfs task to copy index files next to VCFs before running merge
```

tasks/broad/ImputationTasks.wdl
```diff
@@ -382,6 +382,19 @@ task MergeSingleSampleVcfs {
     Int disk_size_gb = 3 * ceil(size(input_vcfs, "GiB") + size(input_vcf_indices, "GiB")) + 20
   }
   command <<<
+    # Move the index file next to the vcf file with the corresponding name
```
gbggrant, line 385:
```suggestion
    # Move the index file next to the vcf with the corresponding name
```
Such nits!

gbggrant (approved):

PR #519, opened 2021-11-22
GL-1737 Adapter Validation Script

wjdingman1:
This PR adds the script from monster to validate that the staging area for the adapter outputs conforms with the DCP/2 spec.

The corresponding hca-adapter-tools PR can be found [here](https://github.com/broadinstitute/hca-adapter-tools/pull/21).


projects/tasks/AdapterTasks.wdl
```diff
@@ -725,4 +729,33 @@ task CreateStagingAreaFile {
...
+    validate-staging-area \
+    --staging-area ~{staging_area} \
+    --ignore-dangling-inputs 
+
+    }
+
+  runtime {
+    docker: docker
+    cpu: cpu
+    memory: "${memory_mb} MiB"
```
gbggrant, line 757: Are we standardized on using the ~{ format?

gbggrant (approved):

nikellepetrillo (approved):

projects/tasks/AdapterTasks.wdl
wjdingman1: ah all the tasks in this WDL are using the ${, should I switch? I have no preference

wjdingman1: according to this I think we're supposed to use $ notation?

https://github.com/broadinstitute/warp/blob/develop/website/docs/contribution/contribute_to_warp/wdl_task_runtime_style.md

PR #529, opened 2021-12-01
Testing Docker Image

jamesemery:

PR #530, opened 2021-12-03
Testing Docker Image baseline for comparison

jamesemery:


jamesemery:
test science please

wjdingman1:
hey @jamesemery, saw you were able to get the test to run and successfully pull the docker images, thats great!

A couple things to mention. The way the tests work is it will automatically test any pipelines that have your changes within their import tree. If you look at the picture below you can see that it tested 6 pipelines, with 2 succeeding and 4 failing. To get more info you can click on the test and view 'Parameters' which will tell you what pipeline was run, then you can see the output via `Console Output` and parse through the logs.

image

In the case of #17244 it ran VariantCalling and the output is shown below. In this instance the test failed because the output VCF from this run differs from the truth set we have. Right now we are just running the 'Plumbing' tests which are significantly scaled down versions of our scientific tests. You can check the actual VCF outputs mentioned in the logs and decide for yourself whether this change is within an acceptable range.

image

If you would like some more help parsing through this or anything please feel free to reach out and we can jump on a call.

@kbergin  or visibility

jamesemery:
Hey @wjdingman1. Thank you for the help. Unfortunately this branch was intended to be a control to isolate changes that might have crept in because of the new GATK version (not counting the other stuff thats getting tested). I'll go digging into what changed but this will somewhat complicate matters...

wjdingman1:
@jamesemery Thats really up to you. Typically when we have changes to tools like GATK, picard etc. we defer to the tool owners to determine whether or not the changes are within an acceptable range. If they are then we just update our truth sets to match the output from pipelines that are using the new tool version

jamesemery:
@wjdingman1 The test failures seem to be related to NaN values in some of that allele specific annotations (the PR in GATK is here: https://github.com/broadinstitute/gatk/pull/7186/files). How feasible in your existing testing framework is to take the output of one of these branches and use it as truth for another branch?

wjdingman1:
@jamesemery That shouldn't be bad but it does take some manual intervention. Just to clarify you want to run the changes you have _against_ the output from GATK 4.2.3 i.e. the 4.2.3 will be the new baseline

jamesemery:
@wjdingman1 It looks like i have a workaround (the change only affects the reblocking code which is not in particular danger of conflict on my main branch). If you look at #529 it looks like all the plumbing tests succeeded. How do i trigger the full scale tests from there?

wjdingman1:
@jamesemery Great news, okay so normally our scientific tests run when we make a PR to Master but here we will have to kick them off manually for each pipeline that had a change.

So looking at your tests we will use #17252 GermlineSingleSample as an example. When you navigate to that test and click `Parameters` you should see the parameters that the test was built with. The important line here is the `ARGS` input, all we will be doing is replacing `-t Plumbing` with `-t Scientific`.

image

To kick off a manual test, click the `warp-workflow-tests` project  tab on the top left of the screen then on the left hand side click `Build with Parameters`. This will bring you to the screen shown below. You will need to select your branch with the changes, and change the test to your desired pipeline. In this case it was `GermlineSingleSample`. Finally you will copy and paste the ARGS from your plumbing tests into the ARGS section here and change the `-t` flag to `Scientific`. The screenshot below should be exactly what you will do for #17252 GermlineSingleSample. 

image

Then click build and note the number of the build so you can come back and check it later. The scientific tests will take a while to run and you will need to repeat this process for all 6 of the pipelines run in your plumbing tests.

To clarify something I said earlier, I misspoke a bit about verifying output. If there are changes in the output we will also verify with the pipeline owners that this change is acceptable, taking the tool owners input into account.

kbergin:
Hi James, if you're seeing changes in the reblocking code we will still need to have those reviewed and approved by @ldgauthier before merging the new GATK version.

jessicaway:
@jamesemery Is this PR still needed?

nikellepetrillo:
closing due to inactivity

PR #531, opened 2021-12-06
GL-1867 - Add RNAWithUmis pipeline to warp test framework.

gbggrant:
Seems to be some non-deterministic metrics outputs?


tests/broad/scala_test/src/main/scala/org/broadinstitute/dsp/pipelines/commandline/ConfigParser.scala
```diff
@@ -910,4 +910,83 @@ class ConfigParser
...
+      opt[Unit]('u', "uncached")
+        .text("Disable call-caching for the main workflow run")
+        .optional()
+        .action { (_, config) =>
+          config.copy(
+            rnaWithUmisConfig =
+              config.rnaWithUmisConfig.copy(useCallCaching = false)
+          )
+        },
+      opt[PapiVersion]("papi-version")
```
jessicaway, line 982: We probably don't need PAPI version anymore since we cannot run on PAPIv1 any longer

verification/VerifyTasks.wdl
```diff
@@ -159,15 +160,37 @@ task CompareBams {
           ~{test_bam} \
           ~{truth_bam} \
           O=comparison.tsv \
-          LENIENT_HEADER=true
-
+          ~{true='LENIENT_HEADER=true' false="LENIENT_HEADER=false" lenient_header}
```
jessicaway, line 163: Can you do `LENIENT_HEADER=~{lenient_header}` or does that fail since `lenient_header` is a boolean rather than a string?

jessicaway:
Looks like the Skylab test failures are happening because this changed the default for `lenient_header` to be `false` in the compare bams task, previously these used `true`

gbggrant:
Thanks @jessicaway I thought I handled my change of the default for CompareBams, but obviously missed all the skylab ones.

tests/broad/scala_test/src/main/scala/org/broadinstitute/dsp/pipelines/commandline/ConfigParser.scala
gbggrant: Oops - too much copying and pasting!

verification/VerifyTasks.wdl
gbggrant: Not sure.  I kind of doubt it, but will try.

jessicaway (approved):

verification/VerifyTasks.wdl
gbggrant: But it did work!  Thanks.

wjdingman1 (approved):

PR #534, opened 2021-12-07
Updated dragmap docker base image from Centos7 to Centos8 

wjdingman1:
This is to patch a security vulnerability

### DIFF
--- a/dockers/broad/dragmap/docker_build.sh
+++ b/dockers/broad/dragmap/docker_build.sh
@@ -2,7 +2,7 @@
 set -e
 
 # Update version when changes to Dockerfile are made
-DOCKER_IMAGE_VERSION=1.0.1
+DOCKER_IMAGE_VERSION=1.1.0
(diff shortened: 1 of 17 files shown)


jessicaway (approved):
Thanks @wjdingman1!

dockers/broad/dragmap/docker_build.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 # Update version when changes to Dockerfile are made
-DOCKER_IMAGE_VERSION=1.0.0
+DOCKER_IMAGE_VERSION=1.0.1
```
gbggrant, line 5: Maybe this should be more like a '1.1.0' version change?

gbggrant (approved):

dockers/broad/dragmap/docker_build.sh
wjdingman1: im cool with whatever, just figured since id go with that since it was a security patch

PR #540, opened 2021-12-09
Update broken libcurl4-ssl to libcurl4-gnutls in VerifyBamID Dockerfile

### LINKED ISSUE #539: verify_bam_id Dockerfile cannot be built anymore
When trying to build the `dockers/broad/verify_bam_id/Dockerfile` the build fails due to an error installing the required packages with the following message:
```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 libcurl4-gnutls-dev : Conflicts: libcurl4-openssl-dev but 7.74.0-1.3+b1 is to be installed
 libcurl4-openssl-dev : Conflicts: libcurl4-gnutls-dev but 7.74.0-1.3+b1 is to be installed
E: Unable to correct problems, you have held broken packages.
```

wjdingman1:
fixes #539

### DIFF
--- a/dockers/broad/verify_bam_id/Dockerfile
+++ b/dockers/broad/verify_bam_id/Dockerfile
@@ -22,7 +22,7 @@ RUN set -eux; \
             gcc \
             git \
             libbz2-dev \
-            libcurl4-openssl-dev \
+            libcurl4-gnutls-dev \
             libhts-dev \
             libssl-dev  \
             unzip \
--- a/dockers/broad/verify_bam_id/README.md
+++ b/dockers/broad/verify_bam_id/README.md
@@ -4,7 +4,7 @@
 
 Copy and paste to pull this image
 
-#### `docker pull us.gcr.io/broad-gotc-prod/verify-bam-id:1.0.0-c1cba76e979904eb69c31520a0d7f5be63c72253-1626442707`
+#### `docker pull us.gcr.io/broad-gotc-prod/verify-bam-id:1.0.1-c1cba76e979904eb69c31520a0d7f5be63c72253-1639071840`
 
 - __What is this image:__ This image is a debian-based image for running the VerifyBamID tool.
 - __What is VerifyBamID:__ VerifyBamId is a tool for DNA contamination estimation from sequence reads using ancestry-agnostic method. See [here](https://github.com/Griffan/VerifyBamID) more information.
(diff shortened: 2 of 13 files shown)


gbggrant (approved):
Looks good. Two questions / kind of open for discussion
1) Should we update the VerifyBamID docker image to have the actual version of the tool in its name?  I think it has a semantic version? There was an Ops discussion about this and I was wondering
2) Do we automatically update the pipelines (as you have here) with the freshly built docker image? I think it's a good idea, but wasn't sure if there's an agreement on that.

jessicaway:
I don't have a strong opinion on (1), but I do think we should automatically use the new docker whenever we build one (point 2) at least as much as possible

jessicaway (approved):

wjdingman1:
Its been the cast that we always update the tasks using the images when the images are update and will roll back/make adjustments if the images throw off the test. 

For part 1 I think that would be good, we have the version pinned to a hash (this was how it was previously, don't know if this was intentional?). Have no problem changing it to a semantic release instead

nikellepetrillo (approved):

michaelgatzen (changes requested):
I have no preference of how to name it, but the version with the current hash has a bug that caused some problems. Both the most current commit hash (6ba8a264552391a2d7d98f38b977c53a01724873) and the latest release (2.0.1) have the fix for that bug

kachulis:
@michaelgatzen but we're not sure the updated version actually fixes the bug so much as randomly avoids it by luck, right?  So it's not clear that updating the version right now just for that is the right choice.  Certainly at least we'd want to do more significant testing to make sure the results from the new version are consistent and reliable before updating.

To bring others up to speed, the bug Michael is talking about is the `-nan` contamination value that we saw recently after thinking it had been fixed a few years ago.  When Michael used the newer version of vbid, the error went away, but we suspect this is similar to when we thought it had been fixed before;  the bug is a super rare edge case, and hitting it is very unstable, so minor unrelated changes can make it look like it has been fixed, even though they aren't fixing the underlying issue, just leading the particular sample being tested to no longer hit the edge case.

wjdingman1:
Okay we are going to put this patch in  for now and then we can address the version change when there is clear answer on how the bug has been addressed. Does that sound reasonable?

michaelgatzen:
Yes sounds good, updating libcurl _really_ should not have an effect on that issue

michaelgatzen (approved):

PR #565, opened 2022-01-10
fixed link to public bucket

ekiernan:
- Added correct link to public Google bucket that hosts the reference panel


khajoue2 (approved):

kayleemathews (approved):
Looks good!

PR #571, opened 2022-01-14
GL-1908. Fix issue with escaping of strings/filenames with spaces embedded

gbggrant:
Fix issue with escaping of strings/filenames with spaces embedded that occurred on older (< 57) versions of Cromwell
t occurred on older (< 57) versions of Cromwell


jessicaway (approved):
Looks good!

kbergin (approved):

PR #577, opened 2022-01-19
Increase Boot Disk for Arrays GATK tasks

gbggrant:
to avoid an out of disk error on WFL's cromwell

### DIFF
--- a/pipelines/broad/arrays/single_sample/Arrays.changelog.md
+++ b/pipelines/broad/arrays/single_sample/Arrays.changelog.md
@@ -1,3 +1,8 @@
+# 2.6.3
+2022-01-18  (Date of Last Commit)
+
+* Increase Boot disk for GATK tasks to avoid an out of disk space error
+
 # 2.6.2
 2022-01-14  (Date of Last Commit)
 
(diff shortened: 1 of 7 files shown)


wjdingman1 (approved):
just curious, why does the size of the boot disk matter?

gbggrant:
@wjdingman1 It's failing [here](https://aou-jmui.gotc-prod.broadinstitute.org/jobs/3ef7e454-cf96-490c-b321-fd1597861344) in AoU/WFL - presumably because it's running an older version of Cromwell than ours (we haven't seen this error in our tests or on non-AoU/WFL launch arrays workflows).  I'm upping the boot disk because I remember running into this issue when I upped a bunch of our workflows to gatk 4.1.8.0 long ago (and at that time I was only upping Germline-related tasks, so the arrrays tasks didn't get the treatment).  At that time, the issue seemed to be that the gatk image itself was too large for the default boot disk on the VM. I'm hopeful that's the issue, but if you see something else / have suggestions I wouldn't mind a little more defensive proramming at this point, as I'm testing on production.

jessicaway (approved):
Looks good!

PR #584, opened 2022-01-19
Fix failing vulnerability scan for imputation images

wjdingman1:
Couple things here:
- Had to do an upgrade of installed packages in the bcf-vcf image so that expat (the failing package) would get patched
- This update caused issues when linking binaries for Minimac
- Switch to using the static binary from Minimac instead so we don't have to worry about compiling it ourselves


gbggrant (approved):

jessicaway (approved):
Thanks @wjdingman1 !

PR #594, opened 2022-01-27
GL-1915. Updated SortSam task to be able to use call caching

gbggrant:
I split SortSam into two different tasks as I didn't like the optional outputs of the existing task.


wjdingman1 (approved):
Much cleaner

khajoue2 (approved):

PR #597, opened 2022-01-31
GL-1919. Fix OOD on Reblocking

gbggrant:


wjdingman1 (approved):

ldgauthier (approved):
2X + 3 to 4X is probably more disk that we need, but these tasks only run for 2 hours at most for genomes (I believe) so it's worth the compute cost to (hopefully) eliminate the manual retry effort cost.

gbggrant:
Yeah, probably overkill, but I'd rather not have to up this again!

khajoue2 (approved):

jessicaway (approved):

PR #602, opened 2022-02-01
Fix error in passing of optional genotypes_index file

gbggrant:
Fix error in passing of optional genotypes_index file to CheckFingerprint task from workflow


jessicaway (approved):
👍

wjdingman1 (approved):

PR #605, opened 2022-02-01
Migrate dragmap base image to rockylinux from centos8

wjdingman1:
Centos 8 just recently went EOL so this is a temporary solution to use [RockyLinux](https://rockylinux.org/) in its place.

I put a ticket on the backlog to circle back around to this if we want to go with debian or alpine.


gbggrant (approved):

jessicaway (approved):
Looks good!

PR #606, opened 2022-02-02
remove site-specific paths from within some outputs to improve reproducibility

mr-c:


gbggrant:
Hi @mr-c could you clarify the intent of this PR? You are stripping out information about the command usage, which could be needed for understanding what has happened in the pipeline.
If you need another output, perhaps you could add that instead?

mr-c:
> Hi @mr-c could you clarify the intent of this PR? You are stripping out information about the command usage, which could be needed for understanding what has happened in the pipeline.

Hello. Yes, I'm happy to give more context. 

The motivation is having bit identical output for the same input files and parameters and software container. Otherwise it is difficult to test for reproducibility.

Additionally, identical outputs for the same inputs reduce costs and time when using systems that do data deduplication and task reuse.

See "Rule 10: Produce identical results when given identical inputs" from Taschuk M, Wilson G (2017) Ten simple rules for making research software more robust. PLoS Comput Biol 13(4): e1005412. https://doi.org/10.1371/journal.pcbi.1005412

For example, over in https://github.com/common-workflow-lab/wdl-cwl-translator we test that the CWL translation of some WDL descriptions from this repository produce the same results from the same inputs as compared to a run captured using the original WDL versions of the descriptions.

I understand why an individual tool might include this information for local, interactive, or otherwise manual running of their tool. But every workflow management system that I know of records provenance information (even if they don't yet export this in a standardized way).  So for the files in this PR (task definitions made to be used from a workflow that will be run on a system that does record such provenance information) I don't see enough benefits to outweigh the costs.

If you are interested in how a workflow management system can present the results of a workflow run and the provenance of the task execution, then I recommend checking out the [Workflow Run RO-Crate profile](https://www.researchobject.org/workflow-run-crate)

> If you need another output, perhaps you could add that instead?

True, I could make the path-free version another output, but for some systems that might double the disk space used. So an inplace edit seems safer.

PR #609, opened 2022-02-07
GL-1859. Refactoring of RNAWithUMIsPipeline

gbggrant:
Mostly whitespace corrections.
Slightly reduced memory and disk usage on several tasks.
Standardized memory sizing

### DIFF
--- a/tasks/broad/RNAWithUMIsTasks.wdl
+++ b/tasks/broad/RNAWithUMIsTasks.wdl
@@ -1,617 +1,628 @@
 version 1.0
 
 task VerifyPipelineInputs {
-    meta {
-        description: "Verify that the pipeline input is either a ubam or pair of fastqs with additional info"
-    }
+  meta {
+    description: "Verify that the pipeline input is either a ubam or pair of fastqs with additional info"
+  }
 
-    input {
-        File? bam
-        File? r1_fastq
-        File? r2_fastq
+  input {
+    File? bam
+    File? r1_fastq
+    File? r2_fastq
 
-        # fastq specific field
-        String? platform
-        String? library_name
-        String? platform_unit
-        String? read_group_name
-        String? sequencing_center = "BI"
+    # fastq specific field
+    String? platform
+    String? library_name
+    String? platform_unit
+    String? read_group_name
+    String? sequencing_center = "BI"
 
-        String docker = "us.gcr.io/broad-dsp-gcr-public/base/python:3.9-debian"
-        Int cpu = 1
-        Int memory_mb = 2000
-        Int disk_size_gb = ceil(size(bam, "GB") + size(r1_fastq,"GB") + size(r2_fastq, "GB")) + 10
-    }
+    String docker = "us.gcr.io/broad-dsp-gcr-public/base/python:3.9-debian"
+    Int cpu = 1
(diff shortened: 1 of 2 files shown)


tasks/broad/RNAWithUMIsTasks.wdl
```diff
@@ -1,617 +1,628 @@
...
-        Int disk_size_gb = ceil(2.2 * size(bam, "GB")) + 300
-    }
+  input {
+    File bam
+    File bam_index
+    String output_bam_basename
+
+    String docker = "us.gcr.io/broad-gotc-prod/umi_tools:1.0.0-1.1.1-1638821470"
+    Int cpu = 2
+    Int memory_mb = 7500
```
wjdingman1, line 577: assuming the original memory allocation was a huge overkill?

wjdingman1 (approved):

tasks/broad/RNAWithUMIsTasks.wdl
gbggrant: Yeah, I looked at these tasks (after running the scientific test) using the output of the memory profiling tool and found I could reduce it significantly

tasks/broad/RNAWithUMIsTasks.wdl
```diff
@@ -1,617 +1,628 @@
...
-        # Note: this is 'us.gcr.io/tag-team-160914/neovax-tag-rnaseq:v1', just pulled into a location visible to warp tests
-        docker : docker
-        cpu : cpu
-        memory : "${memory_mb} MiB"
-        disks : "local-disk ${disk_size_gb} HDD"
-        preemptible: 0
-    }
+    String docker = "us.gcr.io/broad-gotc-prod/samtools-star:1.0.0-1.11-2.7.10a-1642556627"
+    Int cpu = 8
+    Int memory_mb = 64000
```
jessicaway, line 111: 64 Gib seems like a lot here. Should this scale with input size or is a large static amount actually most appropriate?

tasks/broad/RNAWithUMIsTasks.wdl
jessicaway: Was this the monitoring script or some other tool? (just curious)

jessicaway (approved):
Looks good!

gbggrant:
Yeah, it was the monitoring script.  I should double check that number/reduction - now you've got me nervous.

tasks/broad/RNAWithUMIsTasks.wdl
gbggrant: It is a lot. Let me look into making it scale.

tasks/broad/RNAWithUMIsTasks.wdl
gbggrant: Yes, after running the scientific tests, I checked the monitoring log for memory usage.

tasks/broad/RNAWithUMIsTasks.wdl
gbggrant: I've updated this to scale (but I don't think it makes a big difference)

jessicaway: Sounds good!

PR #617, opened 2022-02-14
Update RNAseq pipeline

takutosato:
~~[Still in progress]~~. Ready for review. I have not done the postprocessing of transcriptome bam (we will revisit it after discussing with Junko + Mark).

- [x] Update STAR command line arguments
- [x] Update ribosomal intervals
- ~~Update refFlat~~ (will not do, see comment below)
- [x] Update the exon 1000 bp bed file
- ~~Update docker for Picard CollectRNAMetrics~~ (Not needed)
- ~~Postprocess transcriptome bam~~ (On Hold)


takutosato:
I tried copying the rRNA interval list to gs://broad-references but I don't have permission. Will someone copy for me? It's in:

gs://broad-dsde-methods-takuto/RNA/resources/gencode_v34_UCSC_rRNA.interval_list 

I think it should be renamed to `gencode_v34_rRNA.interval_list`.

No changes to refFlat, gtf, STAR index for now, to make things simple. I'm starting to think including the unlocalized scaffolds is not worth the trouble.

@gbggrant @wjdingman1 I saw that the picard tasks use the docker "us.gcr.io/broad-gotc-prod/picard-cloud:2.26.6." In the draft version of the wdl I shared with you a few months ago, the CollectRNASeqMetrics task used a different jar, that skips the step that makes sure the header of bam and rRNA interval list match, which was useful for development but not fit for production. I have since fixed the problem (i.e. fixed the header of the rRNA interval list), and these tasks can now run with the regular Picard distribution (i.e. does not require my hack fix). I assume this 2.26.6 docker contains my hacky fix, which was never merged into the main branch of Picard---is that right?

gbggrant:
@takutosato I have copied gs://broad-dsde-methods-takuto/RNA/resources/gencode_v34_UCSC_rRNA.interval_list to gs://broad-references/Homo_sapiens_assembly38_noALT_noHLA_noDecoy/v0/annotation/gencode_v34_rRNA.interval_list

As to the picard-2.26.6 docker, that is the standard Picard docker, built on release 2.26.6 of docker. It does NOT include your header 'fix', and has worked fine for us on all of the data we have processed with it.

pipelines/broad/rna_seq/RNAWithUMIsPipeline.changelog.md
```diff
@@ -1,3 +1,12 @@
+# 1.0.2
+2022-02-xx  (Date of Last Commit)
+
+* Updated STAR command line argument
+(TODO) Updated ribosomal intervals
```
kbergin, line 5: Can you clarify a bit more here why you're updating the command line argument or what you're updating it to do now that it didn't do before? And same with the updates to the intervals, reflat etc like if it's that you're moving to a new annotation version just say that and say what the version is. Thanks!

takutosato: Sounds good, will do

gbggrant:
add to whitelist

jessicaway (approved):
looks good!

gbggrant:
retest this please

gbggrant (approved):

PR #630, opened 2022-02-25
GL-1941 - Update to Picard 2.26.11

gbggrant:
Update to Picard 2.26.11
Address obscure bug in GtcToVcf -> VcfToAdpc (some variant metrics, calculated as infinite, were rendered incorrectly in the VCF)
Also got rid of unneeded syntax in Picard usage

### DIFF
--- a/pipelines/broad/arrays/single_sample/Arrays.changelog.md
+++ b/pipelines/broad/arrays/single_sample/Arrays.changelog.md
@@ -1,3 +1,9 @@
+# 2.6.6
+2022-02-025 (Date of Last Commit)
+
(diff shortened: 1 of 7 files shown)


wjdingman1 (approved):
Do we want to update picard everywhere else and other dockers or just limit it to this?

gbggrant:
I'm kind of a mind to just do it in an affected pipeline (here I'm trying to fix a specific bug). It just becomes a bigger job if we do it across the board - and may surface some differences in scientific tests?

jessicaway (approved):

PR #633, opened 2022-02-28
[DFE-720] Wire contamination outputs through wrapper

mmorgantaylor:
testing manually in Terra, will update when complete

### DIFF
--- a/tasks/broad/RNAWithUMIsTasks.wdl
+++ b/tasks/broad/RNAWithUMIsTasks.wdl
@@ -652,6 +652,8 @@ task formatPipelineOutputs {
     String? picard_fingerprint_summary_metrics
     String? picard_fingerprint_detail_metrics
     File unified_metrics
+    Float contamination
+    Float contamination_error
 
     Int cpu = 1
     Int memory_mb = 2000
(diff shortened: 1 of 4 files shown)


takutosato (approved):
Looks great, thank you Morgan. Should I go ahead and merge it?

mmorgantaylor:
sure @takutosato - i'm still testing and might need to make some more changes, but fine with me to merge it to your branch. just not ready for merge to develop yet.

tasks/broad/RNAWithUMIsTasks.wdl
```diff
@@ -690,6 +693,8 @@ task formatPipelineOutputs {
     outputs_dict["picard_quality_distribution_pdf_file"]="~{picard_quality_distribution_pdf}"
     outputs_dict["fp_summary_metrics_file"]="~{picard_fingerprint_summary_metrics}"
     outputs_dict["fp_detail_metrics_file"]="~{picard_fingerprint_detail_metrics}"
+    outputs_dict["contamination"]=str(~{contamination})
```
takutosato, line 696: @mmorgantaylor I was hoping to just use sed within the CalculateContamination task to truncate the digits, but the problem with that approach is sometimes the number is in scientific notation e.g. 0.001 vs 1E-3. It looks like wdl's read_float() automatically converts scientific notation to decimals. So this seems like a good place to truncate the digits. Do you mind adding it to your PR? (I don't have an easy way to test my the warp wdl…)

I think either

`'%.5f'%(float(~{contamination}))`

or 

`'%.5f'%(~{contamination})`

should work, depending on the type of `~{contamination}` in python. Or if you have a different approach please use that, since it seems like you have something working in your branch.

mmorgantaylor: ah yeah sure thing, will add that and test it now!

takutosato: Thanks!

PR #637, opened 2022-03-04
Add "set -e -o pipefail" to imputation tasks

kachulis:
Many imputation tasks involve multiple commands.  Currently, the pipeline can "succeed" even when some of the commands have failed (for example due to running out of memory), as long as the final command in the task returns 0.  This PR fixes this, so that in tasks with multiple commands, the whole task will fail if any command in the task fails.


wjdingman1 (approved):

gbggrant (approved):

jessicaway (approved):

wjdingman1:
add to whitelist

PR #653, opened 2022-03-26
Improve Imputation QC Metrics

kachulis:
The imputation qc metric currently has a very strong and undesirable dependence on cohort size.  This PR fixes the issue by:
1) Switching from using the minimac estimated r2, to the leave-one-out r2.  The leave-one-out r2 is only calculated at genotyped sites, but it is a more accurate evaluative metric, and doesn't have the cohort size related dependence on MAF that causes issues with the estimated r2.
2) evaluate only sites with MAF>0.05 for the metric.  Sites with very low MAF are very difficult to impute, so this restricts the metric evaluation to sites where imputation should perform very well.


wjdingman1:
looks like the plumbing test is failing in `FailQCWellImputedFrac` with the following stderr

`Error: Well imputed fraction was 0.9348135, QC failure threshold was set at 0.95`

I'm guessing this is not expected? You can view the task here if needed -> [gcloud](https://console.cloud.google.com/storage/browser/broad-gotc-dev-cromwell-execution/Imputation/f076b0c6-db55-4d97-b4fe-df27fb691872/call-FailQCWellImputedFrac;tab=objects?authuser=1&prefix=&forceOnObjectsSortingFiltering=false)

wjdingman1:
@kachulis Hey left some comments on this as well don't know if you had seen them

kachulis:
Thanks @wjdingman1!
 I don't think that failure is actually indicative of a problem with the pipeline, but is instead due to the fact that the test data is on a chip with ~600K sites, whereas I have tested this on GDA data, which has ~2M sites.  I think there are three reasonable options to resolve this, and am not sure what I think the best choice is, so would be interested to hear your/other lantern folks' opinions.

1.  We lower the qc threshold for the test to 0.9, as an input in the json.  This has the benefit of leaving the default at something more reasonable for data we actually run in production.
2. We lower the default qc threshold to 0.9.  I don't really like this option, since it isn't really a good choice for our standard production data.  But it does make the default work on a wider range of more common arrays.
3. We get some GDA data for tests.  I think this is probably my preferred choice, since it makes our tests more similar to production.  But it is also likely the largest lift.  We already have data we could use for this, which we've been using this data for clinical validations.  But we would need to get it into the appropriate buckets, etc.

wjdingman1:
I think the easiest patch would be option 1 for now and I think that's fine since it isn't going to affect production runs, but that's not really my call. We could do option 1 with option 3 being a "nice-to-have" and get to it later on down the road.

@kbergin Would you care to chime in on this; I'm sure you have a stronger opinion than I do.

kachulis:
It turns out I was STILL doing this metric calculation slightly wrong.  With that fixed, it now looks like keeping the default threshold at 0.9 makes more sense, so we shouldn't need to make any adjustment for the tests.

nikellepetrillo (approved):

wjdingman1:
retest this please

wjdingman1:
add to whitelist

PR #660, opened 2022-03-31
Take out annotationDB files from per-chromosome JG

ldgauthier:
because they're fat and they were just for development


wjdingman1:
add to whitelist

wjdingman1 (approved):

nikellepetrillo (approved):

ldgauthier:
Awesome, thanks!  Hopefully we don't have to run this anymore, but if we do we can avoid moving around these big files no one cares about.

PR #666, opened 2022-04-07
add new outputs to TDR, update docker tag for ingest script

mmorgantaylor:


tasks/broad/RNAWithUMIsTasks.wdl
```diff
@@ -754,7 +758,7 @@ task formatPipelineOutputs {
   >>>
 
   runtime {
-    docker: "broadinstitute/horsefish:twisttcap_scripts"
+    docker: "broadinstitute/horsefish:tdr_import_v1.1"
```
mmorgantaylor, line 761: this is the same code as before, we just decided to make versioned tags to be safer with updates.

PR #673, opened 2022-04-15
add memory multiplier to sort sam inputs

nikellepetrillo:


wjdingman1 (approved):

tasks/broad/BamProcessing.wdl
```diff
@@ -22,12 +22,16 @@ task SortSam {
...
   # SortSam spills to disk a lot more because we are only store 300000 records in RAM now because its faster for our data so it needs
   # more disk space.  Also it spills to disk in an uncompressed format so we need to account for that with a larger multiplier
   Float sort_sam_disk_multiplier = 3.25
   Int disk_size = ceil(sort_sam_disk_multiplier * size(input_bam, "GiB")) + 20
 
+  Int machine_mem_mb = ceil(5000 * memory_multiplier)
+
+
   command {
     java -Dsamjdk.compression_level=~{compression_level} -Xms4000m -Xmx4500m -jar /usr/picard/picard.jar \
```
gbggrant, line 36: You'll need to scale the -Xms and -Xmx parameters accordingly. I'm sure we have examples of that too.  Typically would set Xmx about 500 less than machine_mem_mb and Xms 1000 less.

pipelines/broad/dna_seq/germline/single_sample/exome/ExomeGermlineSingleSample.changelog.md
```diff
@@ -1,3 +1,8 @@
+# 3.0.7
+2022-04-15 (Date of Last Commit)
+
+* Updated task SortSam in BamProcessing.wdl, this update has no effect on this pipeline
```
gbggrant, line 4: Minor nit - I think BamProcessing is used in all of these affected pipelines, so you might change this comment to say something like 'updated task SortSam in BamProcessing.wdl to take an optional memory_multiplier.
(but I could be wrong that it's not used in the affected pipelines).

gbggrant (changes requested):

gbggrant (approved):

PR #677, opened 2022-04-18
Bash bug fix for how we pass array to the merge scripts

khajoue2:
We passed a bash array with $arr instead of ${arr[@]} so the merge script was not running correctly.


wjdingman1 (approved):
Good catch

nikellepetrillo (approved):

PR #678, opened 2022-04-19
Update dsde-toolbox to patch security vulnerabilities

wjdingman1:
Removed dsde-toolbox from trivy scan and updated it to patch vulnerabilities.


wjdingman1:
@dinvlad Hey kind of weird here, would you have any insight as to why when I scan the image `us.gcr.io/broad-gotc-prod/dsde-toolbox:stable_04-18-2022` it passes fine but there is an issue when being scanned in our action? 

I made sure to have the same severity level, type etc.

dinvlad:
@wjdingman1 sorry for a late reply, is it still an issue?

wjdingman1:
@dinvlad I just removed it from our scan because we are only using the image one time but it was very weird, it was still erroring out

wjdingman1:
@dinvlad was running this in terminal but `trivy image --vuln-type os --ignore-unfixed --no-progress --severity CRITICAL us.gcr.io/broad-gotc-prod/dsde-toolbox:stable_04-18-2021`

dinvlad:
Hmm, yea that's really bizarre I've even tried `docker run --rm -v /var/run/docker.sock:/var/run/docker.sock:ro aquasec/trivy image --vuln-type "os" --severity "CRITICAL" --ignore-unfixed --no-progress --timeout 10m us.gcr.io/broad-gotc-prod/dsde-toolbox:stable_06-10-2021` as in the failing action from 2 days ago. My only guess is maybe the Trivy database wasn't updated back then? 🤷

wjdingman1:
@dinvlad just reintroduced it, looks like its failing

dinvlad:
The one difference I see between Action and local scans is this line
```
DB Repository: ghcr.io/aquasecurity/trivy-db
```
I wonder if that make it somehow behave differently 🤯

dinvlad:
OK, I just did `docker pull aquasec/trivy` and now they show up locally with that [Docker command](https://github.com/broadinstitute/warp/pull/678#issuecomment-1111392260) from above as well! So just a matter of Trivy version, it seems.. Same for the non-Docker version, I suspect.

wjdingman1:
ah that'll do it, thanks for the troubleshoot

wjdingman1:
@dinvlad another question, the scan is also failing because its picking up the AWS Access keys in the image, should this not be ignored with the `--vuln-type` flag? 

wondering if its just a better idea to remove the dsde-toolbox from scanning altogether since its a private image and "insecure" anyway

dinvlad:
Hey @wjdingman1 - good idea to just remove it, if it's not used in deployed production service but only during CI/CD.

wjdingman1:
@khajoue2 can i get a thumb

khajoue2 (approved):

PR #693, opened 2022-04-26
Add vcf index to germline pipelines for new testing framework

wjdingman1:
Added Laura's VCF Index comparisons to the new test framework for the germline pipelines that have been migrated so far


nikellepetrillo (approved):

khajoue2 (approved):
lgtm

PR #711, opened 2022-05-19
Add new timestamp field to populate in TDR

mmorgantaylor:
NOTE: we should not merge this until after the current e2e testing is done, since this relies on a different dataset schema

corresponding changes to docker/script here: https://github.com/broadinstitute/horsefish/pull/96


wjdingman1:
add to whitelist

PR #737, opened 2022-06-08
Wd update wreleaser

wjdingman1:


khajoue2 (approved):
@wjdingman1 Do we need to include the wreleaser binary files in this PR? There are two files under scripts for darwin and linux

wjdingman1:
@khajoue2 yes its good to include them if someone would prefer to use the static binary over having to install go and install tthe package that way (go can be annoying to set up for first time users)

nikellepetrillo (approved):

PR #738, opened 2022-06-08
increasing memory in compareVCFs task

nikellepetrillo:


verification/VerifyTasks.wdl
```diff
@@ -21,7 +21,7 @@ task CompareVcfs {
   runtime {
     docker: "gcr.io/gcp-runtimes/ubuntu_16_0_4:latest"
     disks: "local-disk 50 HDD"
-    memory: "16 GiB"
```
gbggrant, line 24: Could we consider using the memory-doubling retry strategy on this?

gbggrant (approved):

khajoue2 (approved):

verification/VerifyTasks.wdl
ldgauthier: Ultimately I want to deprecate this in favor of my GATK tool, so we don't need to spend too much time boyscouting.

PR #746, opened 2022-06-16
Add CheckFingerprint to scala tests

wjdingman1:
Removed the lod-threshold from test inputs since it was not used and was conflicting with the scala input generator


wjdingman1:
@gbggrant Had to do some name changes here for the new scala tests, basically there is some regex stuff that needs to be done to format inputs for new scala test. 

They will break if the name of the pipeline has a subworklow of the same name i.e. `CheckFingerprint.CheckFingerprint`, hope this change is okay, let me know your thoughts

gbggrant (approved):

nikellepetrillo (approved):

PR #761, opened 2022-07-06
[DFE-770] Support whitespace in sample id strings

mmorgantaylor:
This PR adds support for whitespace in sample id strings, which previously would break TDR ingestion. It goes along with [this PR](https://github.com/broadinstitute/horsefish/pull/100) to make a new version (v1.4) of the ingest-outputs-to-TDR script.


mmorgantaylor:
@meganshand we no longer need the `tdr_sample_id` because the new TDR ingest method we're using, "merge," automatically uses the TDR table's primary key to match the row. so as long as we're including the primary key (`collab_sample_id_run_id`) in the data to ingest, which we do, we don't need to mark it specifically.

if you think that's confusing, though, and that we should more explicitly define the primary key, that's fair and i could refactor. let me know what you think.

meganshand:
Ah, that makes sense. I don't think it's confusing. I think you can leave it as is. Thanks!

nikellepetrillo (approved):

mmorgantaylor:
ah thank you @nikellepetrillo for the reminder about pipeline version! really appreciate it

PR #768, opened 2022-07-14
staging -> master

timaeusx:


nikellepetrillo (approved):

nikellepetrillo:
germlineSingleSample smart test is failing because of the 2 dragen samples. There is some non determinism in the dragen pipeline so we expect these 2 samples to fail.

PR #785, opened 2022-08-03
Updated a check for an empty file produced in the GetFingerprintingIntervalIndices task in JointGenotypingByChromosomePartOne.wdl.

### LINKED ISSUE #767: Check for an empty file in JointGenotypingByChromosomePartOne.wdl needs to be updated.
https://github.com/broadinstitute/warp/blob/baf72dd4adb34c38cf12e818ec4abd7ecce5d105/pipelines/broad/dna_seq/germline/joint_genotyping/by_chromosome/JointGenotypingByChromosomePartOne.wdl#L209

@snovod encountered an issue (in https://broadinstitute.atlassian.net/browse/PO-48542?atlOrigin=eyJpIjoiYTYxOGYyNjhiYWI1NGJlZTliZWRhNTE4NmUxZjFlOGUiLCJwIjoiaiJ9 ) with this condition check, which does not account for the change of behavior in `read_lines` for empty files introduced in https://github.com/broadinstitute/cromwell/pull/6457/. If that change is to be supported, we should instead check that the array is non-empty, and not that its first element is not `””`. The condition can occur when `haplotype_database` intervals and the intervals produced by concatenating the `unpadded_intervals` have no overlap.

samuelklee:
Closes #767.

Note that I also searched for other instances of such a file check in the repo and didn't find any.

### DIFF
--- a/pipelines/broad/dna_seq/germline/joint_genotyping/by_chromosome/JointGenotypingByChromosomePartOne.changelog.md
+++ b/pipelines/broad/dna_seq/germline/joint_genotyping/by_chromosome/JointGenotypingByChromosomePartOne.changelog.md
@@ -1,3 +1,8 @@
+# 1.4.8
+2023-04-05 (Date of Last Commit)
+
+* Updated a check for an empty file produced in the GetFingerprintingIntervalIndices task
+
 # 1.4.7
 2022-11-04 (Date of Last Commit)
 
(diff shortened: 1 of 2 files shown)


kbergin (approved):
LGTM but I added eng from the team to review as well.

jessicaway:
add to whitelist

snovod (approved):

jessicaway (approved):

samuelklee:
Apologies @snovod and others, clearing out the backlog during the retreat and just noticed that this never got merged. I think I need to update the changelog, but will then go ahead and merge afterwards.

samuelklee:
OK, I've updated the version/changelog. Not sure if this was actually necessary. Also not sure if I'm the one who should merge the PR---feel free to let me know if this is the case or to go ahead and merge otherwise. Sorry, new to this repo!

samuelklee:
@snovod I've updated the version and changelog. If all looks good, please merge on my behalf!

PR #790, opened 2022-08-09
moving scientific ownership to sam lee instead of laura

nikellepetrillo:


khajoue2 (approved):

.pullapprove.yml
```diff
@@ -160,7 +164,7 @@ groups:
       request: 2
     reviewers:
       users:
-        - ldgauthier # Laura Gauthier
+        - samuelklee # Samuel Lee
```
samuelklee, line 167: I'm happy to help review any changes to these pipelines, but strictly speaking, I think Laura only handed off responsibilities for the joint-genotyping pipelines to me. Perhaps @kachulis can clarify who the primary owner for the single-sample pipelines will be going forward?

samuelklee: (Or if it turns out that I do own these now, it would be good to sort that out!)

samuelklee: OK---chatted with Chris and I understand the intended split now. He'll own everything up through alignment/duplicate marking.

jessicaway (approved):

.pullapprove.yml
nikellepetrillo: thanks sam! that sounds good 👍

PR #791, opened 2022-08-09
Update BuildIntervalList to use samtools docker image

timaeusx:


khajoue2 (approved):

pipelines/skylab/build_indices/BuildIndices.wdl
```diff
@@ -355,7 +355,7 @@ task BuildIntervalList {
   >>>
 
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-umitools:0.0.1"
+    docker: "us.gcr.io/broad-gotc-prod/samtools:1.0.0-1.11-1624651616"
```
jessicaway, line 358: Let's rollback this change so that we don't publish a new version of the pipeline that doesn't run

pipelines/skylab/build_indices/BuildIndices.changelog.md
```diff
@@ -1,3 +1,7 @@
+# 1.0.1
```
jessicaway, line 1: Let's rollback this change so that we don't publish a new version of the pipeline that doesn't run

jessicaway (approved):

PR #804, opened 2022-08-24
Adding filtering to Ultima Joint Genotyping pipeline

meganshand:
Here is a draft of the Ultima Joint Genotyping pipeline that adds filtering. This needs a release of GATK to remove some test docker tags before being merged. I'm hoping for some initial review before the next GATK release though. 

Tests will also currently fail since the expected outputs need to be updated.


tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+
+    source ~/.bashrc
+    conda activate genomics.py3
+
+    python -m tarfile -e ~{ref_fasta_sdf} ~{ref_fasta}.sdf
+
+    run_comparison_pipeline.py \
+            --n_parts 0 \
+            --hpol_filter_length_dist 12 10 \
+            --input_prefix $(echo "~{input_vcf}" | sed 's/\(.vcf.gz\|.vcf\)$//') \
```
gileshall, line 248: For readability, I would consider breaking out the subshell and assigning its output to a local variable.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+    String docker = "us.gcr.io/broad-dsde-methods/broad-gatk-snapshots/gatk-remote-builds:mshand-05a76349aac401cc5d08ad0375c45fa9d4c7c864-4.2.6.1-48-g05a76349a"
+  }
+
+  command <<<
+    set -eo pipefail
+    
+    tail -n +2 ~{thresholds} > ~{thresholds}.body
+    cat ~{thresholds}.body | awk 'BEGIN { FS=","} \
+    {print "-filter \"VARIANT_TYPE == \047" $1 "\047 && ~{score_key} \
+     < "$2 "\" --filter-name LOW_SCORE_"$1 " "}' | tr '\n' ' ' > tmpfile
```
gileshall, line 326: I can't decide if this is fragile or not.  What leapt out at me is how long I was staring at this, trying to unravel how this works.  I don't think this needs to be rewritten, but in case it does become fragile, I suggest adding a comment or two to help unpack its intent.

gileshall (commented):
Awesome, looks great!  I added a few minor suggestions, but overall, I say ship it.

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/UltimaGenomicsJointGenotyping.changelog.md
```diff
@@ -1,3 +1,8 @@
+# 1.1.0
+2022-08-24 (Date of Last Commit)
+
+* Adding filtering to UltimaGenomicsJointGenotyping pipeline. Output VCFs are now filtered.
```
samuelklee, line 4: Adding -> Added to follow the changelog style guide.

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/UltimaGenomicsJointGenotyping.wdl
```diff
@@ -34,6 +39,23 @@ workflow UltimaGenomicsJointGenotyping {
     # than a z-score of -4.5 which is a p-value of 3.4e-06, which phred-scaled is 54.69
     Float excess_het_threshold = 54.69
 
+    #inputs for threshold filtering
+    String sample_name_truth
+    File gtr_vcf
```
samuelklee, line 44: Is `gtr_vcf`, etc. conventional? Why not just `truth_`?

samuelklee: Note also the use of `truth_sample_name` and `call_sample_name` in the `run_comparison_pipeline.py` command line, so it might be worth just mirroring those names as well.

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/test_inputs/Plumbing/plumbing.inputs.json
```diff
@@ -15,5 +15,18 @@
...
-"UltimaGenomicsJointGenotyping.ImportGVCFs.gatk_docker":"us.gcr.io/broad-dsde-methods/broad-gatk-snapshots:UG_feature_branch_v4"
+"UltimaGenomicsJointGenotyping.ImportGVCFs.gatk_docker":"us.gcr.io/broad-dsde-methods/broad-gatk-snapshots:UG_feature_branch_v4",
+"UltimaGenomicsJointGenotyping.snp_annotations": "-A ReadPosRankSum -A FS -A SOR -A QD -A AVERAGE_TREE_SCORE -A AVERAGE_ASSEMBLED_HAPS -A AVERAGE_FILTERED_HAPS",
+"UltimaGenomicsJointGenotyping.indel_annotations": "-A MQRankSum -A ReadPosRankSum -A FS -A SOR -A QD -A AVERAGE_TREE_SCORE",
+"UltimaGenomicsJointGenotyping.flow_order": "TGCA",
+"UltimaGenomicsJointGenotyping.ref_fasta_sdf": "gs://broad-gotc-test-storage/tmp/jbx/reference_sdf.tar",
+"UltimaGenomicsJointGenotyping.runs_file": "gs://broad-gotc-test-storage/tmp/jbx/runs.conservative.bed",
+"UltimaGenomicsJointGenotyping.annotation_intervals": ["gs://broad-gotc-test-storage/tmp/jbx/LCR-hs38.bed", "gs://broad-gotc-test-storage/tmp/jbx/mappability.0.bed", "gs://broad-gotc-test-storage/tmp/jbx/exome.twist.bed"],
+"UltimaGenomicsJointGenotyping.use_allele_specific_annotations": true,
```
samuelklee, line 25: Should you be using `AS_` annotations above?

Incidentally, I wonder if we should fail if AS annotations are specified but the names do not start with `AS_`? This seems a bit restrictive, since I don't know that this naming convention is actually documented or enforced.

Also, should we go back and add this argument to the canonical WDL? What will be the strategy for keeping canonical WDLs in the GATK repo in sync with the ones here?

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/test_inputs/Scientific/scientific.inputs.json
```diff
@@ -15,5 +15,18 @@
...
-"UltimaGenomicsJointGenotyping.ImportGVCFs.gatk_docker":"us.gcr.io/broad-dsde-methods/broad-gatk-snapshots:UG_feature_branch_v4"
+"UltimaGenomicsJointGenotyping.ImportGVCFs.gatk_docker":"us.gcr.io/broad-dsde-methods/broad-gatk-snapshots:UG_feature_branch_v4",
+"UltimaGenomicsJointGenotyping.snp_annotations": "-A ReadPosRankSum -A FS -A SOR -A QD -A AVERAGE_TREE_SCORE -A AVERAGE_ASSEMBLED_HAPS -A AVERAGE_FILTERED_HAPS",
+"UltimaGenomicsJointGenotyping.indel_annotations": "-A MQRankSum -A ReadPosRankSum -A FS -A SOR -A QD -A AVERAGE_TREE_SCORE",
+"UltimaGenomicsJointGenotyping.flow_order": "TGCA",
+"UltimaGenomicsJointGenotyping.ref_fasta_sdf": "gs://broad-gotc-test-storage/tmp/jbx/reference_sdf.tar",
+"UltimaGenomicsJointGenotyping.runs_file": "gs://broad-gotc-test-storage/tmp/jbx/runs.conservative.bed",
+"UltimaGenomicsJointGenotyping.annotation_intervals": ["gs://broad-gotc-test-storage/tmp/jbx/LCR-hs38.bed", "gs://broad-gotc-test-storage/tmp/jbx/mappability.0.bed", "gs://broad-gotc-test-storage/tmp/jbx/exome.twist.bed"],
+"UltimaGenomicsJointGenotyping.use_allele_specific_annotations": true,
```
samuelklee, line 25: Same comment here.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+workflow ExtractOptimizeSingleSample { 
+    input {
+        Array[File] input_vcf
+        Array[File] input_vcf_index
+        String base_file_name
+        String sample_name_calls
+        File gtr_vcf
+        File gtr_vcf_index
+        File gtr_highconf_intervals
+        String sample_name_gtr
```
samuelklee, line 16: Same comment about `gtr` here.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+        String docker = "gcr.io/terra-project-249020/jukebox_vc:test_jc_optimize_3d7509"
+    }
+    String output_vcf = basename(input_vcf, ".vcf.gz") + ".~{sample_name}.vcf.gz"
+    command <<<
+        set -eo pipefail
+        source ~/.bashrc
+        conda activate genomics.py3
+        bcftools view -s ~{sample_name} ~{input_vcf} -o  tmp.vcf.gz -O z 
+        bcftools index -t tmp.vcf.gz
+        bcftools view -h tmp.vcf.gz | sed 's/COMBINED_TREE_SCORE,Number=A,/COMBINED_TREE_SCORE,Number=\.,/g' > hdr.fixed.txt
```
samuelklee, line 130: Parts of this WDL seem like they could be generically useful, no? It might be worth thinking about exposing arguments needed in this Ultima-specific use case and moving towards generic tasks/sub-WDLs at some point, if not now.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+
+  command <<<
+    set -e
+
+    source ~/.bashrc
+    conda activate genomics.py3
+
+    python -m tarfile -e ~{ref_fasta_sdf} ~{ref_fasta}.sdf
+
+    run_comparison_pipeline.py \
```
samuelklee, line 245: Perhaps even specific tasks for calculating metrics or performing comparisons could work through a more generic interface. Maybe a catch-all argument for specific arguments like the flow order? Again, OK for future work.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
samuelklee: Is there some WDL pattern using e.g. `basename` for doing this? (Maybe not, since there are two possible suffixes?)

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+            --n_parts 0 \
+            --hpol_filter_length_dist 12 10 \
+            --input_prefix $(echo "~{input_vcf}" | sed 's/\(.vcf.gz\|.vcf\)$//') \
+            --output_file ~{input_vcf_name}.comp.h5 \
+            --gtr_vcf ~{gtr_vcf} \
```
samuelklee, line 250: Hmm OK, I guess `gtr` appears here.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+    output {
+        File output_vcf = "~{output_vcf_name}"
+        File output_vcf_index = "~{output_vcf_name}.tbi"
+        File monitoring_log = "monitoring.log"
+    }
+}
+
+task CompareToGroundTruth {
+  input {
+    String left_sample_name
```
samuelklee, line 213: Why `left` and `right` here?

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
samuelklee: Yes, let's comment and perhaps simplify. I don't think I've unraveled it myself!

For example, not sure I understand the presence of `LOW_SCORE_`. Is this trying to construct the filter name used by ScoreVariantAnnotations? If so, note that the filter name there is also an exposed argument, and it's not necessarily given by `LOW_` + `score_key`; it's just that the default key and filter names are `SCORE` and `LOW_SCORE`, respectively.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,388 @@
...
+        disks: "local-disk " + disk_size + " HDD"
+        docker: docker
+    }
+
+    output {
+        File output_vcf_file = "~{output_basename}.vcf.gz"
+        File output_vcf_index = "~{output_basename}.vcf.gz.tbi"
+    }
+}
+
```
samuelklee, line 386: Extra whitespace.

samuelklee (approved):
Thanks @meganshand, looks great! Just a few minor comments and questions.

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/UltimaGenomicsJointGenotyping.wdl
samuelklee: And never mind, I see `gtr` elsewhere below. But in general, it might be good to mirror arguments when possible

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/test_inputs/Plumbing/plumbing.inputs.json
meganshand: Good catch, we should be using `AS_` annotations. I think failing on non `AS_` annotations is too restrictive. Especially if we want to use a combination of `AS_` and non `AS_` annotations (which should be ok I think). Maybe a warning, but honestly I probably wouldn't have noticed a warning in this case either.

For keeping the WDLs in sync: That is an excellent question. @jessicaway A version of this WDL also currently lives in GATK. It's good to have it in GATK so that it can be tested as GATK tools change rather than waiting for a GATK release to test it, but we also want it available as a production pipeline. Do you have any ideas on what the process should be for where these WDLs belong and/or how they should be tested? 

I am looking back at a previous slack conversation with @kbergin on this subject and our conclusion was to put it in GATK first and then "cross that bridge when we come to it". I think we're at the bridge now though :)

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
meganshand: Thank you for commenting this because I realized that this line is outdated and is no longer needed. I made the task more generic, but I'm not sure if you're suggesting that it live somewhere else more generic as well? Should I move it to Utilities.wdl? Every time I touch anything in Utilities I have to update the versions of every pipeline that uses them which seems like a minor drawback.

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/test_inputs/Plumbing/plumbing.inputs.json
jessicaway: @meganshand this is a very good question and something we have struggled within the past. There is definitely a need for integration testing in our pipelines as GATK is updated. We often find that things look different when we go to update versions of GATK (or Picard for that matter) and it is a painful process to capture all the changes and track down which commit caused the difference(s). 

That being said, keeping the WDLs in 2 locations that do not sync automatically will almost certainly not be a sustainable solution. 

I have not spent much time looking into a solution, but off the top of my head, my first idea is to add a test that runs on GATK PRs (maybe PRs to master?) that checks out WARP and runs the pipeline test with the new code. That is not as straightforward as it sounds since we would need to do work to change out the docker used in each task to point to the new version of GATK. 

We are definitely open to ideas and suggestions for how to make this better :)

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
meganshand: I attempted a rewrite of this that will hopefully be clearer (with an additional comment too). 

The `LOW_SCORE_` is trying to construct a filter name of the type of variant (it's broken up by snp, hmer indel, nonhmer indel, etc.) which is not clear at all unless you know what's in the `thresholds` file. Maybe it makes more sense to name the filter `"LOW_" + name_of_score_key + "_" + variant_type`, but maybe it's cleaner to hard code the filter name as `"LOW_SCORE_" + variant_type`? It's convenient/confusing that the annotation happens to be `SCORE` in this case.

Either way, I'm testing it now and will push the changes on Monday once it works.

pipelines/broad/dna_seq/germline/joint_genotyping/UltimaGenomics/test_inputs/Plumbing/plumbing.inputs.json
meganshand: I think this is an ideal solution, but would probably take some effort to set up in GATK. I think that's worth doing, but I'm also not sure what to do in the meantime with something like this Ultima JG WDL. To complicate the matter it's only one subworkflow of the whole pipeline that currently resides in GATK. 

Is there a way to have WARP pull the pipeline from GATK in the short term before we get tests set up in GATK that check out WARP? Or is that an overly complicated solution to a temporary problem (although to be fair, I'm not sure how temporary)?

It might be worth talking this through off of this PR with all relevant parties.

meganshand:
So sorry for the delayed response here. I believe I've addressed the comments on this PR and tried to fix all the tests. The outstanding issues that I need some help with are:

- [ ] Updating the truth data for the Ultima Joint Genotyping pipeline tests. We expect the output to change drastically with this PR since we're adding filtering. I want to make sure this is updated correctly so I don't break the tests on develop/main.
- [ ] Have a plan for how to handle WDLs that belong in either GATK or WARP. As discussed with @jessicaway I'll set up a meeting to talk through this part in person.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
```diff
@@ -0,0 +1,401 @@
...
+        String base_file_name
+        File input_vcf
+        File input_vcf_index
+        String docker = "gcr.io/terra-project-249020/jukebox_vc:test_jc_optimize_3d7509"
+    }
+
+    String output_vcf_name = "~{base_file_name}" + ".vcf.gz"
+    command <<<
+        set -e
+        source ~/.bashrc
```
jessicaway, line 183: just wondering why this line is necessary?

jessicaway (approved):
Looks good! A couple minor comments/questions

tasks/broad/JointGenotypingTasks.wdl
```diff
@@ -1077,4 +1077,41 @@ task PartitionSampleNameMap {
...
+}
+
+# Task used to divide annotations that were summed by GenomicsDB to get average values.
+task CalculateAverageAnnotations {
+  input {
+    File vcf_index
+    File vcf
+    Array[String] annotations_to_divide = ["ASSEMBLED_HAPS", "FILTERED_HAPS", "TREE_SCORE"]
+
+    String docker = "us.gcr.io/broad-dsde-methods/broad-gatk-snapshots:UG_feature_branch_v4"
```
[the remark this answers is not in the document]
meganshand: It is as of this week! However, this is the docker we've been using in all of the Ultima pipelines. It seems like it might be better to update them all at once in a separate PR? Happy to hear if you think I should update it here first.

tasks/broad/UltimaGenomicsGermlineFilteringThreshold.wdl
meganshand: huh, the `.bashrc` sets up conda, so I think it's needed for `conda activate genomics.py3`, but this task (and some of the others) don't use python so I think it can be removed. I'll clean this up.

meganshand: actually I was wrong, in this docker gatk and bcftools are managed by conda. I'll add a comment in the code.

meganshand:
Retest this please

nikellepetrillo (approved):

PR #832, opened 2022-09-23
added disk memory and cpu to each task as input

khajoue2:


nikellepetrillo:
looks good, but it does look like the scATAC test is failing with a difference in some metrics https://gotc-jenkins.dsp-techops.broadinstitute.org/job/warp-workflow-tests/34884/console i'm not sure if that is expected

nikellepetrillo (approved):

timaeusx (approved):
Looks good!

PR #835, opened 2022-09-27
Fix for large Ultima samples in Single Sample pipeline

meganshand:
We've seen that large samples (80X coverage for example) run out of disk on the `MakeOptionalOutputBam` step. All that that task is doing is creating an optional File type for downstream use. It's expensive to run (it localizes the bam only to return the optional file type) so this PR folds that functionality into the previous task (MarkDuplicatesSpark). Now MarkDuplicatesSpark generates an optional output bam.

This should not change any outputs, in other words both the Germline and Somatic Ultima pipeline tests should pass without updating the truth.

### DIFF
--- a/tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl
+++ b/tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl
@@ -268,12 +268,16 @@ task MarkDuplicatesSpark {
   input {
     Array[File] input_bams
     String output_bam_basename
+    Boolean save_bam_file
     String docker = "us.gcr.io/broad-dsde-methods/broad-gatk-snapshots:UG_feature_branch_v4"
     Int disk_size_gb
     Int cpu = 32
     Int memory_mb = if 4 * ceil(size(input_bams, "MB")) / 4000 > 600000 then 300000 else 208000
     Int preemptible = 0
   }
+
+  String dummy_file_name_for_optional_output_bam = if save_bam_file then output_bam_basename + ".bam" else "path_that_does_not_exist"
+  String dummy_file_name_for_optional_output_bai = if save_bam_file then output_bam_basename + ".bam.bai" else "path_that_does_not_exist"
 
   parameter_meta {
     input_bams: {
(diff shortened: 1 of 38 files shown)


meganshand:
Tests are now passing so I think this is ready for review.

khajoue2 (approved):

tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl
```diff
@@ -268,13 +268,17 @@ task MarkDuplicatesSpark {
...
     String output_bam_basename
+    Boolean save_bam_file
     String docker = "us.gcr.io/broad-dsde-methods/broad-gatk-snapshots:UG_feature_branch_v4"
     Int disk_size_gb
     Int cpu = 32
     Int memory_mb = if 4 * ceil(size(input_bams, "MB")) / 4000 > 600000 then 300000 else 208000
     Int preemptible = 0
   }
 
+  String dummy_file_name_for_optional_output_bam = if save_bam_file then output_bam_basename + ".bam" else "path_that_does_not_exist"
```
jessicaway, line 279: This is an interesting way to do this. This seems like a silly question, but have you tried this with `save_bam_file` set to false? Seems like the behavior could be a little strange (returning a string "path_that_does_not_exist" as a file)

jessicaway (approved):
One minor question, otherwise looks good!

tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl
meganshand: Yeah I think we did add a test where `save_bam_file` is false. That would be tested by the `UltimaCramOnly` pipeline. The thing I'm trying to get here is a path that doesn't exist since that is how to create an undefined value for an optional File type. The spec seems to imply this will work in this section: https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#files-and-optional-outputs

This would fail if there is somehow a file on the VM with the path "path_that_does_not_exist" either from the script or from the docker. Hopefully this is unlikely enough that it's relatively safe. If there is a more standard way to not fill in optional outputs I'm happy to change this to match.

PR #840, opened 2022-10-11
Force fragment size histogram file to exist in rnaseqc output

kachulis:
If there are no fragments used to generate fragment size histogram, rnaseqc does not generate a fragment size histogram file, which causes the task to fail.  This PR forces an empty file to be produced in this case, so that the task succeeds.

### DIFF
--- a/pipelines/broad/rna_seq/RNAWithUMIsPipeline.changelog.md
+++ b/pipelines/broad/rna_seq/RNAWithUMIsPipeline.changelog.md
@@ -1,3 +1,8 @@
+#1.0.10
+2022-10-11 (Date of Last Commit)
+
+* Force task rnaseqc2 to produce an empty fragment size file when rnaseqc2 does not produce this file due to insufficient data.
+
 # 1.0.9
 2022-09-27 (Date of Last Commit)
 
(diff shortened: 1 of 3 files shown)


mmorgantaylor (approved):

jessicaway:
add to whitelist

jessicaway (approved):

pipelines/broad/rna_seq/RNAWithUMIsPipeline.changelog.md
```diff
@@ -1,3 +1,8 @@
+#1.0.10
```
jessicaway, line 1: you need to update the `pipeline_version` variable in the WDL as well:
https://github.com/broadinstitute/warp/blob/develop/pipelines/broad/rna_seq/RNAWithUMIsPipeline.wdl#L23

jessicaway:
`BroadInternalRNAWithUMIs.changelog.md` also needs to be updated. (No need to increment the version, just add a bullet point to the notes for the most recent version)

jessicaway:
retest this please

jessicaway:
you can ignore the failing trivy check, it is not related to these changes

Issue #842, opened 2022-10-14
rnaseqc undefined behavior for bias metrics in tests.


kachulis:
There is a bug in rnaseqc which can lead to undefined behavior when calculating bias metrics for highly downsampled data (such as our test data).  When only 1 gene is used in the bias calculations, it reads data passed the end of vector, which leads to non deterministic results, which cause tests to fail when not call caching.  This is not particularly worrisome for production because  this bug only occurs when only 1 gene is used for the bias calculation, which should not happen for good production data (and which rnaseqc informs us of in its metrics file).

This should still probably be fixed, either by:

1. fixing the bug in rnaseqc
2. removing rnaseqc from the pipeline
3. adjusting our test data to not hit this edge case

kachulis:
see for example smart test failures on 51b10cf34e82219a2735d7f3596ab2d71ccd36d2

nikellepetrillo:
Hi Chris, 

Is removing rnaseqc really an option? Would we need to replace it with something else? 

-Nikelle

kevinpalis:
Hi @kachulis, is this still an issue?

kachulis:
Nope, this was fixed by #869

PR #849, opened 2022-10-27
Update rnaseqc to address edge case in bias metric calculations

kachulis:
Addresses #842.  Issue was fixed in getzlab/rnaseqc@c426536132b7f90e108707c5e4da206097e31ad7


kachulis:
Combining with #869, so closing.

PR #851, opened 2022-11-04
Update GATK docker in Ultima pipelines

meganshand:
This PR updates the GATK dockers in the Ultima pipelines (CramOnly, WholeGenomeGermline, and JointGenotyping) to the release 4.3.0.0. The differences between the current version and 4.3.0.0 include multiple known improvements that change the outputs including a bug fix to the Allele Filtering code in HaplotypeCaller and a fix to an edge case bug in the HAPDOM annotation. This means we expect the outputs to change for `UltimaGenomicsWholeGenomeGermline` tests. I've looked over the differences and made two additional changes to the pipeline to ensure that our results are now better and that our pipeline aligns with what Ultima runs internally:

- add the UG high quality regions to the single sample filtering step (this change is not as impactful as the following one)
- change the threshold which outputs variants in the transformation from GVCF -> VCF which is then used in the single sample random forest filtering step. We are changing this threshold from 0 (let all the variants through into the random forest model) to 30 (removes low quality variants from the random forest model). 

A side note about that second change: in the current pipeline changing this threshold effectively makes sites with QUAL<30 uncallable since in the ReblockGVCFs step we remove all sites with TREE_SCORE<.2. By not putting QUAL<30 sites into the Random Forest they never get a TREE_SCORE which GATK effectively treats as a score of 0. I'm leaving this as is for now because the results of keeping that TREE_SCORE<.2 threshold are better in joint callsets.

All this to say that I think it's ok to update the truth data for the Single Sample Ultima pipelines. I don't expect other tests to fail so will look into any other difference that pop up.

Here are the sensitivity and precision values for the Scientific truth sample we currently use (which is NA12878 so we have truth data):

image

image


samuelklee (approved):
Seems straightforward enough, LGTM! Thanks for the detailed PR description as well.

One more general question, for my education—what are the WARP conventions for updating workflow versions when underlying tasks are updated? I guess I want to understand if it’s expected that when I run the same version of a workflow, I should always expect identical results (no matter when I grab that workflow from the repo), or if the only guarantee is that the WDLs for the workflow and any subworkflows have not been changed (modulo changes in imported tasks).

meganshand:
I still need to update the workflow versions here! They should absolutely be updated if any underlying tasks changed. That particular test here should fail (and tell me which wdl versions need updating) and then I'll fix that before merging. Sorry for the confusion, I should probably figure out how to run that test before making the PR in the future rather than fix it up after making the PR each time.

meganshand:
As I was writing the update for the JointGenotyping pipeline I realized that I only updated the JointGenotyping tasks that the Ultima pipeline uses (ImportGVCFs and GenotypeGVCFs). Should we update the GATK version across all of the JointGenotyping tasks to keep them in sync (for example GnarlyGenotyper)? Not sure if this is a question for @samuelklee or @jessicaway.

meganshand:
Based on a slack conversation with @jessicaway I'm trying to keep the GATK versions in the repo in sync. This latest commit updates all gatk:4.2.6.1 to gatk:4.3.0.0. I'll see what tests break and if it looks ok to keep the updated GATK version I'll go through and update the changelogs/version numbers on the WDLs at that point.

kachulis:
@meganshand @jessicaway the scientific test failures Megan pointed me to in the RNAWithUMIs pipeline are all known  non-determinism causing negligible differences.  So not a blocker to merging.

timaeusx:
test this please

jessicaway (approved):

gbggrant (approved):

PR #892, opened 2022-12-20
Np fix jg crosscheck input

nikellepetrillo:


jessicaway (approved):

verification/VerifyJointGenotyping.wdl
```diff
@@ -57,10 +57,12 @@ workflow VerifyJointGenotyping {
       truth_text_files = truth_intervals
   }
 
-  call CompareFingerprints {
-    input:
-      test_fingerprint = test_fingerprint,
-      truth_fingerprint = truth_fingerprint
+  if (defined(test_fingerprint)){
```
samuelklee, line 60: Does this assume that `truth_fingerprint` is always defined when `test_fingerprint` is?

samuelklee (approved):
Just one minor question, otherwise LGTM!

verification/VerifyJointGenotyping.wdl
nikellepetrillo: Hi @samuelklee , good question. 

If `truth_fingerprint` was not defined, this task would fail on line 64 because `truth_fingerprint = select_first([truth_fingerprint])` would not be able to be evaluated. However, this should never be the case: `test_fingerprint` and `truth_fingerprint` will either be defined together or neither will be defined. 

The test/truth fingerprint comes from the [TestJointGenotpying.wdl](https://github.com/broadinstitute/warp/pull/892/files#diff-61eebb54b0fd64839c53f5540af1d236e236b223e8fd9060516022ecbf8dbe84R187). That GetFingerprint task will only run if `cross_check_fingerprints = true` and will always output both the truth/test fingerprint.

nikellepetrillo:
retest this please

timaeusx (approved):
LGTM!

nikellepetrillo:
Just to note: BroadInternalRNAWithUMIS and RNAWithUMIS are failing because of suspected non-determinism.

PR #897, opened 2023-01-06
Update gatk docker 

jessicaway:
There were a couple remaining GATK dockers to update after the recent effort


nikellepetrillo (approved):
thanks jess!

jessicaway:
Adding Eng only label since scientific owners were already consulted on the GATK version change

PR #902, opened 2023-01-23
address mb vs gb inconsistency

jessicaway:
### Description

----
The memory values labeled `machine_mem_mb` in LoomUtils.wdl were actually being specified in GiB

### Checklist 

### DIFF
--- a/tasks/skylab/LoomUtils.wdl
+++ b/tasks/skylab/LoomUtils.wdl
@@ -17,7 +17,7 @@ task SmartSeq2LoomOutput {
     String pipeline_version
     Int preemptible = 3
     Int disk = 200
-    Int machine_mem_mb = 18000
+    Int machine_mem_mb = 16000
     Int cpu = 4
   }
 
@@ -89,7 +89,7 @@ task OptimusLoomGeneration {
 
     Int preemptible = 3
     Int disk = 200
-    Int machine_mem_mb = 18000
+    Int machine_mem_mb = 16000
     Int cpu = 4
   }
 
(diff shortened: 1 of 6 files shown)


ekiernan (approved):

nikellepetrillo (approved):

khajoue2 (approved):

tasks/skylab/LoomUtils.wdl
```diff
@@ -17,7 +17,7 @@ task SmartSeq2LoomOutput {
     String pipeline_version
     Int preemptible = 3
     Int disk = 200
-    Int machine_mem_mb = 18
```
khajoue2, line 20: 18000 sounds like an odd value to request! I wonder if it was a typo

tasks/skylab/LoomUtils.wdl
```diff
@@ -89,7 +89,7 @@ task OptimusLoomGeneration {
 
     Int preemptible = 3
     Int disk = 200
-    Int machine_mem_mb = 18
```
khajoue2, line 92: would it make sense to make it 16000?

tasks/skylab/LoomUtils.wdl
jessicaway: Not sure. Happy to adjust it if you think another value is more suitable as a default

tasks/skylab/LoomUtils.wdl
jessicaway: Sure. Do you want me to change all instance of 18GB in this WDL to 16000MB?

khajoue2: I think it makes more sense to be 16000. Thank you!

tasks/skylab/LoomUtils.wdl
khajoue2: Let's change it to 16000. I believe we will be charged for the machine that fits the cpu and memory and 18G is not standard. It should be either 16G or 32G

Issue #772, opened 2022-07-18
Container for DropletUtils using 'root' folder as PATH for scripts


BioNica:
Dear developers,

we are trying to run Optimus locally on a shared environment (HPC, non-root user)
In this environment we use **Singularity** instead of Docker to run the containers.
For most containers called by the Optimus pipeline, the local images are working fine.
We just were not able to port the DropletUtils container called by RunEmptyDrops.wdl

The reason appears to be the placement of the scripts at the `root` directory during container build

https://github.com/broadinstitute/warp/tree/develop/dockers/skylab/emptydrops
https://github.com/broadinstitute/warp/blob/develop/dockers/skylab/emptydrops/Dockerfile

`WORKDIR /root/ `

[...]

```
# Add EmptyDrop scripts
RUN mkdir /root/tools
COPY emptyDropsWrapper ./tools/emptyDropsWrapper
COPY npz2rds ./tools/npz2rds

ENV PATH="/root/tools/emptyDropsWrapper/:/root/tools/npz2rds/:$PATH" 

```

For the compatibility with Singularity and shared environment, would it be possible to change the scripts location to `/opt `or some place accessible by non-root users?

Also changing the working directory to the default local user entry point would avoid permission issues.

Thank you!

kevinpalis:
This is done and merged via PR: https://github.com/broadinstitute/warp/pull/904

PR #904, opened 2023-01-23
Kp update emptydrops container pd 2131

kevinpalis:
### Description

The changes here address concerns outlined in https://github.com/broadinstitute/warp/issues/772 - avoiding the usage of the root directory inside the container. In addition, a few optimizations were added:

- Moved image to GCR instead of Quay
- Conformed to (most) of our docker style guideline
- Build time decreased to 1 hour from 1.5-2 hours
- Image size reduced to 1.5GB from 3GB

----

### Checklist 

**No functionality changes. WDL is only updated to use the new container which should be functionally identical to the old one.**


khajoue2 (approved):
Looks good.

kevinpalis:
Currently, smart tests are failing because of a missing R package. The way emptyDrops handles R dependencies is via a separate script which was failing silently before. I modified it to stop and report errors and now we see this:

`#9 3132.6 package 'optparse' is not available (for R version 3.5.1)`

which is the same package missing in the smart test error messages. Unfortunately, we cannot upgrade to a newer version of R because apparently, this thing depends on DropUtils which only run on 3.5 (according to the initial developer - written on inline comments). I am currently looking into how we can install this package in the build. Otherwise, we'll have no choice but to re-write the entire R wrapper OR revert to the old image in Quay (which will mean we are stuck in that version and cannot fix the #772 or anything related to this container)

kevinpalis:
Looks like smart tests now pass. A few changes that made that possible:

- R upgraded to version 4.2.2 (from 3.5)
- BiocManager upgraded to version 3.16 (from 3.8)
- Added installation of libpcre2-dev to the Dockerfile (as a prerequisite of R 4.x)
- Added a line to manually load BiocParallel package to the emptyDropsWrapper.R (apparently with previous versions this one was automatically loaded, but not anymore)

PR #933, opened 2023-02-28
Kp improving scitests error reporting pd 2130

kevinpalis:
### Description

Currently, the compilation of test/truth differences is a manual process. For every dataset whose generated test data differs from the expected truth data, the differences must be compiled by hand and shared with the scientific owner. This PR will add a task for VerifyMetrics workflow that will consolidate all error messages of failed metrics tests into one file: failed_metrics_file.txt - removing the need to go shard by shard finding all failed metrics. 

You can find an extensive document on this pain point here: https://broadworkbench.atlassian.net/wiki/spaces/PD/pages/2669346855/Improvements+to+Scientific+Test+Validation+Reports

----


kevinpalis:
@ekiernan - I'm not sure if the checklist also applies to our test workflows but I'm tagging you just in case. While this doesn't modify any input/output, it does add a task to VerifyMetrics, which I believe is something only used in our test suites.

verification/VerifyMetrics.wdl
```diff
@@ -74,16 +83,48 @@ task CompareMetricFiles {
       --INPUT ~{file1} \
       --INPUT ~{file2} \
       --OUTPUT ~{output_file} \
-      ~{true="--METRICS_TO_IGNORE" false="" length(metrics_to_ignore) > 0} ~{default="" sep=" --METRICS_TO_IGNORE " metrics_to_ignore}
+      ~{true="--METRICS_TO_IGNORE" false="" length(metrics_to_ignore) > 0} ~{default="" sep=" --METRICS_TO_IGNORE " metrics_to_ignore};
```
nikellepetrillo, line 86: do we need the semicolon at the end?

nikellepetrillo (approved):
lgtm! thank you

verification/VerifyMetrics.wdl
kevinpalis: Not really. It's an oversight on my part, but it allows for additional commands on that block. Originally I was doing the collection of error messages in this task, after the semicolon. But later on decided to write a separate task for it. I'm hesitant to remove it though cause it will trigger another round of tests and approval but won't affect any functionality. Do you prefer I remove it?

nikellepetrillo: nope! fine with me to leave!

jessicaway (approved):
this looks good, thanks @kevinpalis !

PR #937, opened 2023-03-03
Disk size calculation for  SplitMultiSampleVcf

kachulis:
The disk size calculation for `ImputationTasks.SplitMultiSampleVcf` does not scale correctly.  It works for smaller numbers of samples simply because there is is enough buffer built in.  However, we have already seen it fail at ~400 samples.

We want to be able to predict the size of a single sample vcf split out from a multisample vcf with N samples.  I propose the simple size model of size= $(N + n_{overhead})b$ where N is the number of samples in the vcf, $n_{overhead}$ represents the number of samples that the overhead in the file corresponds to (mostly, the site level information for each vcf record).  As can be seen below, this model is imperfect, but good enough to use to get a reasonable estimate of required disk space. 

image
 

By comparing the size of vcfs with multiple samples to the size of a single sample vcf, we find an estimate of $n_{overhead}$ at various sizes.

image

Based on this, I decided to use an $n_{overhead}$ of 20, which give us a reasonable amount of buffer.


tasks/broad/ImputationTasks.wdl
```diff
@@ -873,11 +873,12 @@ task FindSitesUniqueToFileTwoOnly {
 task SplitMultiSampleVcf {
  input {
     File multiSampleVcf
+    Int nSamples
 
     String bcftools_docker = "us.gcr.io/broad-gotc-prod/imputation-bcf-vcf:1.0.7-1.10.2-0.1.16-1669908889"
     Int cpu = 1
     Int memory_mb = 8000
-    Int disk_size_gb = ceil(3*size(multiSampleVcf, "GiB")) + 100
+    Int disk_size_gb = ceil(21*nSamples*size(multiSampleVcf, "GiB")/(nSamples+20)) + 100
```
rickymagner, line 881: This line is a bit mysterious without any context. Maybe link to this PR with explanation on the formula for future readers?

rickymagner (approved):
Looks good, with a small suggestion to add some explanation for the disk_size formula.

schaluva (approved):

tasks/broad/ImputationTasks.wdl
jessicaway: Or just a code comment with the formula (or a note that this is not random)

jessicaway (approved):

nikellepetrillo (approved):

PR #958, opened 2023-04-07
rnaseqc bias edge case patch

kachulis:
getzlab/rnaseqc#79 can lead to very large output metric value which can cause problem for downstream ETL.  Patched here.


mmorgantaylor (approved):
assuming the test passes, looks good

jessicaway (approved):

kachulis:
test this please

PR #962, opened 2023-04-12
Moved whitelist parameters as input files

vsmalladi:
### Description

White list was hardcoded as gs paths. Made them defaults and moved to inputs for workflow so that they will be able to be inputed for different paths for azure. 

----

### Checklist 

### DIFF
--- a/pipelines/skylab/optimus/Optimus.wdl
+++ b/pipelines/skylab/optimus/Optimus.wdl
@@ -36,6 +36,11 @@ workflow Optimus {
     Int tenx_chemistry_version
     # Whitelist is selected based on the input tenx_chemistry_version
     File whitelist = checkOptimusInput.whitelist_out
+    
+    
+    # 10x parameters
+    File whitelist_v2 = "gs://gcp-public-data--broad-references/RNA/resources/737k-august-2016.txt"
+    File whitelist_v3 = "gs://gcp-public-data--broad-references/RNA/resources/3M-febrary-2018.txt"
 
     # Emptydrops lower cutoff
     Int emptydrops_lower = 100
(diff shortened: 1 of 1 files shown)


kevinpalis (approved):
LGTM!

jessicaway:
Hi @vsmalladi, 

I believe you are trying to make the pipeline flexible to accept whitelist files from sources other than GCP, is that correct?

The pipeline should already allow this. 

The reason the defaults are listed outside the inputs section is to have the pipeline automatically select the whitelist based on the chemistry. This behavior and defaults can be overridden by explicitly providing a whitelist as input to the pipeline. Here is the input that allows this: https://github.com/broadinstitute/warp/blob/develop/pipelines/skylab/optimus/Optimus.wdl#L38

Does that address the concern you are trying to address with this PR or is there a further issue?

vsmalladi:
> Hi @vsmalladi,
> 
> I believe you are trying to make the pipeline flexible to accept whitelist files from sources other than GCP, is that correct?
> 
> The pipeline should already allow this.
> 
> The reason the defaults are listed outside the inputs section is to have the pipeline automatically select the whitelist based on the chemistry. This behavior and defaults can be overridden by explicitly providing a whitelist as input to the pipeline. Here is the input that allows this: https://github.com/broadinstitute/warp/blob/develop/pipelines/skylab/optimus/Optimus.wdl#L38
> 
> Does that address the concern you are trying to address with this PR or is there a further issue?

@jessicaway  I get that, but then this assumes that the person knows about the files before trying to run the pipeline. Moving it to Inputs allows the files to be known by the user and UI in terra to have a user put in Azure or other location for the reference data.

vsmalladi:
@jessicaway yes lets do a call next week will reach out on slack.

jessicaway:
This Pr does not appear to contain a meaningful change at this point and we have addressed this issue in separate ways

PR #977, opened 2023-05-11
GroupByUMIs memory accessible at top level workflow

kachulis:


kachulis:
@jessicaway I think it's because this uses a subworkflow.  It seems like (at least in the Terra UI), task input are exposed, but it doesn't propagate further than one level (ie, the inputs to the tasks called in the subworkflow don't seem to be accessible).  Maybe they can be set via json, and this is just a Terra UI limitation?  Although @mmorgantaylor I believe you said you couldn't set this parameter with wfl without this patch?

If there is a better way to do this, I would be thrilled to not have to clutter up the inputs for things like this.

kachulis:
@mmorgantaylor sounds good, let's hold off on merging this until you confirm is it necessary.

jessicaway:
For future reference (this was discussed in a slack thread), it appears that one of the workflows in the chain did not include the `allowNestedInputs = true` meta parameter, which is why Terra was not exposing inputs for this task as optional workflow inputs.

nikellepetrillo:
@mmorgantaylor do we still need this pr? thanks!

kachulis:
this is no longer necessary after #989

PR #982, opened 2023-05-16
Updating buildIndices inputs 

nikellepetrillo:
### Description

We built new references for the BICAN group, so we need to update our input jsons to reflect what we used (which gtf/genome assembly) to create the reference bundles for BICAN.

----

### Checklist 


ekiernan (approved):
The JSONs look good. I have never run the Refseq refs with our Biotypes.tsv, so curious if we're going to see any issues there.

When we build, we'll want to confirm that we're using the docker with the latest modify-gtf and add-introns script updates. 

We'll also want to verify that outputs of the build-indices have some of the selected biotypes and have no PAR genes.

kevinpalis (approved):
LGTM!

PR #985, opened 2023-05-18
Kp moving containers from warp pd 2185

kevinpalis:
### Description

This cleans up all the containers from WARP, deprecates ones that we don't use anymore, and moves everything else to warp-tools. The corresponding branch in warp-tools is here: https://github.com/broadinstitute/warp-tools/tree/kp_moving_containers_from_warp_PD-2185


kevinpalis:
Accidentally added changes that are not supposed to be part of this branch. Deleting this PR and making a new one later on.

PR #1011, opened 2023-06-13
removing alt contigs and chrM from BGE calling interval list

meganshand:
We discovered that there are alt contigs and chrM in the BGE calling interval list v1. This adds a step to the script that will generate a new version of the intervals (v1.1) that removes these extra contigs so that there are no calls in final callsets with alt contig calls.

@jessicaway Is there a better way to keep the provenance of the v1 intervals? Or is the git history enough for that?

@kachulis Could you also please review?

Once this script is settled we'll also need to update the files in the Broad references directory.


jessicaway:
@meganshand as you said, once we finalize the script and generate the new interval list, we will push the new list to the references. There should be a readme living next to the old interval list and describing how that one was generated. We can edit the readme to include the new list and describe what changed. Does that seem like sufficient provenance to you?

jessicaway (approved):
This looks good to me. The only thing I notice is that the output is being incremented from v1 to v1.1. I'm guessing that is to indicate that it is not a new set of intervals, just a patch to the version 1 set. Make sense if that's the case, just wanted to confirm this shouldn't be v2 (or some other version)

meganshand:
Yeah I think that's right (v1.1 vs v2), but happy to hear if @kachulis has a different opinion on the version number.

meganshand:
I think updating the README makes sense. I'll do that and change the version numbers in the script so that they are all 1.1 (I think I left one v1 in there which is now an intermediate file).

kachulis:
I agree v1.1 seems right, not v2

scripts/generate_bge_exome_calling_regions.sh
```diff
@@ -32,7 +32,7 @@ rm tmp.interval_list
 # =============== mirbase.v22.interval_list ================
 # ==========================================================
 
-wget -nc -O mirbase.v22.gff https://www.mirbase.org/ftp/22/genomes/hsa.gff3
+wget -nc --no-check-certificate -O mirbase.v22.gff https://www.mirbase.org/ftp/22/genomes/hsa.gff3
```
kachulis, line 35: was the `--no-check-certificate` necessary?  if it's not needed, it does make me a little nervous, tbh.

scripts/generate_bge_exome_calling_regions.sh
```diff
@@ -107,6 +107,17 @@ java -jar $picard IntervalListTools -UNIQUE \
 
 #Produced 296381 intervals totalling 165239993 bases.
 
+#Limit intervals to chr1-22,chrX,chrY (remove alts and chrM)
```
kachulis, line 110: have we confirmed with MPG and GP folks that they don't have objections to removing chrM, since it was included in the old exome calling regions?  (I don't expect objections, but we should make sure they are aware)

kachulis (approved):
Looks good, just would like to make sure to loop in MPG/GP folks before using in production.

scripts/generate_bge_exome_calling_regions.sh
meganshand: I think I only needed this due to the environment I was running in. I reran on my laptop and didn't need this argument and then double checked that the outputs matched and everything looked good.

scripts/generate_bge_exome_calling_regions.sh
meganshand: Ah, I didn't realize they were also in the old calling regions. One conclusion here could be to just leave them in. It's possible you'd still get homoplasmic alt calls (although they'll be output as diploid `1/1`). 

Alternatively, @kachulis who from the GP side do you think is the right person to reach out to about this?

@ldgauthier Is an ATGU meeting the right place to get MPG buy in that it's ok to remove chrM from the exome calling interval list or do you think documenting it through GP's process will likely be enough since this should be a non-controversial issue?

kachulis: I think the same group we talked to about the initial changes.  Maybe just follow up in that same email thread?

meganshand: The only comment from that email thread was that we should document the change so I'm comfortable removing chrM from the interval list.

@jessicaway I'll be sure to write up and a share a blurb with you for the readme that will document this change.

meganshand:
@jessicaway I think this PR is ready to be merged, but it might require another reviewer? Can you please recommend someone?

nikellepetrillo (approved):

jessicaway:
@meganshand you're all set. Please open a PO ticket for adding these to the public references. There are instructions here: https://broadworkbench.atlassian.net/wiki/spaces/PD/pages/2667184176/How+to+submit+a+request+to+add+reference+files+to+the+Google+public+bucket
You can just provide what to include in a README update

meganshand:
@jessicaway Here's the ticket for copying the file: https://broadinstitute.atlassian.net/browse/PO-62593 

Let me know if you need anything else! I'm assuming you'll be the one to push the merge button on this PR?

jessicaway:
We're happy to merge for you 👍 For future reference, you are also welcome to merge

PR #1014, opened 2023-06-16
[DO NOT MERGE] Create an alignment only version of the Exome Germline Single Sample WDL

jessicaway:
### Description

This branch is in response to a user request for a version of the ExomeGermlineSingleSample pipeline that only does alignment (not variant calling). 

THIS WILL NOT BE MERGED

This is for demonstration and testing purposes only.
----


jessicaway:
the exome test passes on this branch. There are issues with the reprocessing WDLs and tests, but these are ignored, since this is not intended to be merged in.

PR #1081, opened 2023-09-18
ReblockGVCFs now takes GVCFs that are not in the same location as their index

meganshand:
This change allows you to run ReblockGVCFs with input GVCFs in different locations than the index files (for example, if the data is in TDR).


meganshand:
@jessicaway This is the change to ReblockGVCFs that will let GP run it themselves on data from TDR. Tests should pass (this change shouldn't change any outputs) so I think this is ready for review.

ldgauthier (approved):

jessicaway (approved):
looks good 👍

nikellepetrillo (approved):

PR #1088, opened 2023-09-29
KM Update metrics descriptions

kayleemathews:
### Description

This PR is to update the Optimus metrics doc and fix issues raised in Issue https://github.com/broadinstitute/warp/issues/1079.

----

### Checklist 


ekiernan:
Thanks for reviewing this - do you think we should specify in the Program column that it's TagSort instead of WARP-tools? We could take them directly to the code at this link: https://github.com/broadinstitute/warp-tools/tree/develop/tools/TagSort.

nikellepetrillo (approved):
looks good!!

kayleemathews:
@nikellepetrillo and @ekiernan Thank you for taking a look at this. 

> Thanks for reviewing this - do you think we should specify in the Program column that it's TagSort instead of WARP-tools? We could take them directly to the code at this link: https://github.com/broadinstitute/warp-tools/tree/develop/tools/TagSort.

@ekiernan Yes I think that's a good idea. I'll make that change.

> Also wonder if it would be helpful to define cell barcode and UMI - we've had folks in recent meetings kind of use them interchangeably, but here they have distinct definitions.

@ekiernan I didn't realize this was something people were confused about. If that's the case, then yes, let's definitely do that.

ekiernan (approved):
Thank you so much for updating this doc!

PR #1137, opened 2023-11-29
Fixes ValidateVcfs to handle when VCF is not colocated with index

meganshand:
I missed that the `ValidateVCF` task has the input VCF as an input in the ReblockGVCFs pipeline if the optional `calling_interval_list` is not provided. This should finally make the entire ReblockGVCFs pipeline robust to VCFs in separate locations from their index files (such as in TDR).


jessicaway (approved):

kevinpalis (approved):
LGTM!

khajoue2 (approved):

PR #1153, opened 2023-12-18
updating gatk to 4.5.0.0

meganshand:
This updates most of the GATK versions used in warp to GATK 4.5.0.0. I skipped BQSR for now because the results changed which was unexpected. Those can be found in `BamProcessing.wdl`: `BaseRecalibrator`, `ApplyBQSR`, and `GatherBQSRReports`. 

The other changes that were found from the tests are:

- Ultima single sample pipelines (`BroadInternalUltimaPipeline` and `UltimaGermlineSingleSample`): The results change here because there is randomness in the single sample filtering strategy that is applied (`TREE_SCORE`) and therefore all the results change. This is expected. I checked the scientific tests to make sure nothing was becoming egregiously worse and overall the F1s look ever so slightly better with GATK 4.5.0.0. The precision is slightly worse, but the sensitivity is better and all within very small differences that I'm not concerned with.

image

- For the `ReblockGVCFs` and `VariantCalling` pipelines, the update changes one line in the VCF headers.

```
< ##INFO=<ID=RAW_GT_COUNT,Number=3,Type=Integer,Description="Counts of genotypes w.r.t. the reference allele in the following order: 0/0, 0/*, */*, i.e. all alts lumped together; for use in calculating excess heterozygosity">
---
> ##INFO=<ID=RAW_GT_COUNT,Number=3,Type=Integer,Description="Counts of genotypes w.r.t. the reference allele: 0/0, 0/*, */*, i.e. all alts lumped together; for use in calculating excess heterozygosity">


jessicaway (approved):

kayleemathews:
Looks like this might need some additional documentation updates since I last looked. I can take a look this afternoon.

kayleemathews (approved):
Documentation is updated!

PR #1216, opened 2024-02-27
Update JointGenotyping.inputs.json

ldgauthier:
Wrong file name in WGS joint genotyping intervals prevents scatters wider than ~210

### Checklist 

- [x] Did you modify, delete or move: file paths, file names, input names, output names, or task names?

@ekiernan


jessicaway:
add to whitelist

nikellepetrillo:
@ldgauthier @rickymagner Do you still want this PR merged in? Is this the interval list that should be used in our WARP tests?

ldgauthier:
Yes please!  I believe it's the same territory, but split wider, so tests shouldn't change, but it would like to use this for tests as well.

PR #1251, opened 2024-03-26
ReblockGVCFs requires less memory for WGS samples when no interval list is provided

meganshand:
### Description

When a calling interval list is not provided to ReblockGVCFs, ValidateVariants takes a great deal of memory on WGS samples. When the WGS sample was processed with DRAGEN, there is no calling interval list, but DRAGEN skips some regions of the genome, so the reblocked GVCF does not validate unless it is using the input GVCF as the interval list. 

We fix the need for a large amount of memory in this case by first converting the GVCF input into an interval list (where abutting intervals are merged together). In the future, GATK might be able to solve this issue internally (see https://github.com/broadinstitute/gatk/issues/8608), but for now this workaround should fix the memory issue.

----

### Checklist 

- [x] Did you add inputs, outputs, or tasks to a workflow?


kayleemathews (approved):
Docs are updated!

nikellepetrillo (approved):
Looks good! i kicked of the scientific reblock tests off of this branch and they succeeded 🎉 https://gotc-jenkins.dsp-techops.broadinstitute.org/job/warp-workflow-tests/71142/console

jessicaway (approved):

meganshand:
Thank you! @jessicaway Do I need one more review before this can be merged?

PR #1257, opened 2024-04-01
Updating picard version in QC.wdl to avoid error with updated Platform Name (PL) in the SAM/BAM/CRAM file.

jiveshs98:
### Description

_Updating picard docker version in QC.wdl._
- Changed the docker value from us.gcr.io/broad-gotc-prod/picard-cloud:2.26.10 to us.gcr.io/broad-gotc-prod/picard-cloud:3.1.1.
- I noticed that the task ValidateSamFile failed for CRAMs that contained "ELEMENT" as PL value. However, this shouldn't be the case as "ELEMENT" was added as a recognized value in the SAM format specification in August 2022. (See Appendix B: SAM Version History at [here](https://samtools.github.io/hts-specs/SAMv1.pdf) .
- So, proposing to update the docker version of picard cloud to 3.1.1 because it is the latest tag and was created after the aforementioned change to the SAM format specification. This would help to prevent unnecessary pipeline failures from QC.WDL.


nikellepetrillo:
Hi @jiveshs98, unfortunately, our team does not have the bandwidth to update Picard to 3.1.1.

PR #1313, opened 2024-07-01
updating gatk to version 4.6.0.0

meganshand:
There are changes in the most recent GATK release that I'd like for the BGE pipeline, so this is a draft PR that changes it everywhere from 4.5.0.0 -> 4.6.0.0 so that I can track broken tests in a centralized place (which we expect for at least the JointGenotyping pipelines).


meganshand:
retest this please

meganshand:
All Scientific test differences are now accounted for! The changes are addressed here: https://github.com/broadinstitute/gatk/pull/8973

The major differences are:

In the ExomeGermlineSingleSample, ExomeReprocessing, JointGenotyping, ReblockGvcf, WholeGenomeGermlineSingleSample, and WholeGenomeRepprocessing pipelines we see the following expected differences:

- Expected 0/0 -> ./. (no call genotype) changes in ReblockGVCFs and JointCalling. This also changes the AN annotation.
- HaplotypeCaller drops some low quality variant calls. This occurs before Reblocking, so the only difference we see in the ReblockedGVCF (that already dropped this site) is a change in DP for the genotype.
- If AD is 0 for non-ref reads (which happens with DRAGEN inputs) then AS_QD will have jitter because the AS_QD is calculated to be infinity. These sites have expected AS_QD differences (that are frequently quite different from QD since that doesn’t use AD for the depth calculation)
- Dragen mode pipelines have small differences from both dragmap non-determinism and small changes to HaplotypeCaller at low quality sites.

In the UltimaGenomicsJointGenotyping and UltimaGenomicsWholeGenomeGermline pipelines we see the following expected differences:

- Bug fix to UltimaJointGenotyping pipeline. Some sites had the wrong number of values in the allele specific annotations, but this is now fixed.
- Ultima single sample pipeline has some expected differences in the single sample calls from a HaplotypeCaller change.

Therefore the Scientific tests that will need to be updated are the following:
- BroadInternalUltimaGenomics
- ExomeGermlineSingleSample
- ExomeReprocessing
- JointGenotyping
- ReblockGvcf
- UltimaGenomicsJointGenotyping
- UltimaGenomicsWholeGenomeGermline
- WholeGenomeGermlineSingleSample
- WholeGenomeRepprocessing

I will take a stab at updating the change logs to reflect at least the pipelines with significant changes and clean up a few things and then this will be ready for review.

PR #1362, opened 2024-08-20
Lk pd 2736 doublets

ekiernan:
This PR adds new dockers for warp-tools and for merge-star-npz in order to accomodate new calculations for doublet scores. 

The warp-tools docker has been modified to include a new script for calculating doublet scores using a modified DoubletFinder from the Allen Institute. 

It then uses this script to calculate new library-level metrics such as percent doublets and new keeper cell metrics. 

The merge-npz-docker has been modified with a new combined_matrix.py script that removes the calculation of the keeper cell metrics so that it can be calculated after doublet scores.

The Optimus WDL no longer calculates keeper metrics in the mergestaroutputs task, but instead does this calculation during h5ad production using h5adUtils.wdl.

This PR also updates documentation.


ekiernan:
Smart-test failures are as expected. We added two new columns to the h5ad. Gene metric comparisons fail because we've made gene_names unique now. When we make truth gene_names unique the metrics match.

ekiernan:
retest this please

jessicaway (approved):
Looks great! Thanks @ekiernan !

ekiernan:
First run of smart-tests here: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/warp-smart-tests/8500/

ekiernan:
The only smart-test that failed was scATAC - which ran because we modified the VerifyTasks.

PR #1400, opened 2024-10-24
Reducing memory required for ValidateVCF in Reblock pipeline

meganshand:
In the ReblockGVCFs pipeline, there is a step to validate the reblocked GVCF. If the user does not provide an interval list to validate over, then the pipeline uses the input GVCF itself as the interval list to validate over. This frequently occurs with DRAGEN whole genome GVCFs since there isn't an associated interval list, and DRAGEN drops some regions so a generic WGS interval list cannot be used. 

By adding the argument --VARIANT_ID_METHOD USE_FIRST to the command, the method in htsjdk that combines adjacent intervals [requires much less memory](https://github.com/samtools/htsjdk/pull/1711).


nikellepetrillo (approved):

aawdeh (approved):

jessicaway (approved):

nikellepetrillo:
retest this please

PR #1410, opened 2024-11-04
Lk fix parse barcodes and join barcodes

ekiernan:
This PR fixes a bug where the bgzipped tsv is being overwritten with an empty TSV in join barcodes (Multiome) and in parse barcodes (PairedTag)

### DIFF
--- a/tasks/skylab/H5adUtils.wdl
+++ b/tasks/skylab/H5adUtils.wdl
@@ -255,12 +255,13 @@ task JoinMultiomeBarcodes {
     set -e pipefail
 
     # decompress the bgzipped fragment file
-    echo "Printing the atac_fragment_base"
-    echo "~{atac_fragment_base}"
+    echo "Moving fragment file for bgzipping"
+    mv ~{atac_fragment} /cromwell_root
+    echo "Listing files, should see .gz fragment file"
+    ls -l
     echo "Decompressing fragment file"
-    bgzip -d ~{atac_fragment}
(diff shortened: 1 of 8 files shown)


tasks/skylab/H5adUtils.wdl
```diff
@@ -255,16 +255,18 @@ task JoinMultiomeBarcodes {
     set -e pipefail
 
     # decompress the bgzipped fragment file
+    echo "Moving fragment file for bgzipping"
```
jessicaway, line 258: Wondering why this is necessary? I've seen pushback against using "cromwell_root" before (can't recall if it was an issue for different clouds or HPC environments). Since you're moving things into the directory explicitly here it seems like it should be robust, assuming the runner has permission to write to "cromwell_root"

jessicaway (approved):

tasks/skylab/H5adUtils.wdl
ekiernan: It was necessary because it would not bgzip without being moved to working directory and it would not pipe to a file. I even tested it in the docker and could not get bgzip to output a separate file.

ekiernan:
Testing Multiome Scientific here: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/warp-workflow-tests/88070/

Testing PairedTag Scientific here: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/warp-workflow-tests/88071/

PR #1428, opened 2024-11-15
Np starsolo fails silently when it runs out of disk

nikellepetrillo:
### Description

While we were optimizing STARsolo for slide-tags, we discovered a bug where STARsolo fails silently and produces a truncated BAM file. 
The BAM sizes for the 3 shards we used were truncated ~19 GB, 25 GB, and 25 GB and had a flag in the BAM header indicating an EOF (end of file) error. 
This header caused a failure during the MergeBAM task, the task that follows STARSolo. The STARSolo logs indicated no problem with STARsolo running. 

This PR adds bam validation step to the StarSolo task so it will catch the EOF error and fail before continuing onto the MergeBam task

----

### Checklist 


ekiernan (approved):
LGTM! Thank you!

pipelines/skylab/optimus/Optimus.wdl
```diff
@@ -93,6 +93,7 @@ workflow Optimus {
   String star_docker = "star:1.0.1-2.7.11a-1692706072"
```
peterhuene, line 93: Sorry for the drive-by comment after the PR was merged:

This declaration appears to be no longer used after the changes in this PR (this was caught by our CI process for `sprocket` where we analyze public WDL repositories for changes to diagnostics emitted by sprocket).

PR #1439, opened 2024-11-29
Regenie4 noqc

shengqh:
allow qc_pgen files as input of regenie4


shengqh:
sorry that I misplaced it. Would you please delete it? Thanks.

PR #1443, opened 2024-12-03
Update output names for ReblockGVCF

jessicaway:
### Description

Change requested by the variants team

----

### Checklist 

- [X] Did you modify, delete or move: file paths, file names, input names, output names, or task names?


pipelines/broad/dna_seq/germline/single_sample/ugwgs/UltimaGenomicsWholeGenomeGermline.wdl
```diff
@@ -202,8 +202,8 @@ workflow UltimaGenomicsWholeGenomeGermline {
 
   # Outputs that will be retained when execution is complete
   output {
-    File output_gvcf = ReblockGVCF.output_vcf
-    File output_gvcf_index = ReblockGVCF.output_vcf_index
+    File output_gvcf = ReblockGVCF.reblocked_gvcf
```
koncheto-broad, line 205: Do we need to change the outputs here to? I don't think we (GVS) ever direct consume the output of this WDL, so it's likely that maintaining its output naming scheme was deliberate.  But I just wanted to verify that here

jessicaway: This change is necessary since the output of the reblock workflow is being referenced. I left the output name for this workflow the same, but needed to update the reference to the output of ReblockGVCF

koncheto-broad: yup, totally makes sense why this change was needed.  As I don't know the users of this particular wdl, I mostly wanted to ensure that it wasn't something any end users would be running before sending the data into GVS (in which case we'd want to update its outputs as well).  It sounds as though this is not expected behavior though, and this one has only been touched because it consumes ReblockGVCF for other reasons.  Thanks!

jessicaway: It's a good question. I'm not sure whether we expect the outputs of this workflow to be used in joint calling. 
@rickymagner would you be able to comment on whether we expect the outputs of UltimaGenomicsWholeGenomeGermline to be used for joint calling in GVS?

rickymagner: Hm I don't know if GVS has been used on Ultima samples before. Does anyone from the Variants team know if that's an existing/planned application for GVS?

jessicaway: Thanks @rickymagner! 
Sounds like this isn't currently a known issue @koncheto-broad, let me know if this becomes an issue in the future and we can change it here too.

koncheto-broad (approved):
LGTM

pipelines/broad/dna_seq/germline/single_sample/ugwgs/UltimaGenomicsWholeGenomeGermline.wdl
koncheto-broad: Thanks for clarifying, y'all!  I think we pretty exclusively use Illumina samples at this point, so the Ultima pipeline is unlikely to affect us.  And if we discover that it does, we can always just reach out then!

nikellepetrillo (approved):

jessicaway:
retest this please

ekiernan (approved):
LGTM!

PR #1468, opened 2025-01-07
TSPS-341 remove tasks for recovering variants not in the reference panel

jsotobroad:
### Description

The workflow was updated to remove any tasks used only for recovering variants that are in the input that do not overlap with the reference panel. step 2 in this [doc](https://docs.google.com/document/d/15HQFRPc0tCdueXLFFF28jPYCUic50XJojEK72J81Cq0/edit?pli=1&tab=t.0)

We also separated out beagle imputation specific tasks into its own file and kept the common imputation tasks in the ImputationTasks.wdl file

Both Imputation and Beagle Imputation wdls validate.

We also tested that the  difference in outputs between pre and post these changes as we expect sites to exist in pre changes that do not exist in post changes

run on chr20
pre changes (truth.vcf.gz) - https://bvdp-saturn-dev.appspot.com/#workspaces/teaspoons-imputation-dev/teaspoons_imputation_dev_control_workspace_20240726/job_history/71ab7fe2-98bc-41db-a271-dd944c8ae643

post changes (test.vcf.gz) - https://bvdp-saturn-dev.appspot.com/#workspaces/teaspoons-imputation-dev/teaspoons_imputation_dev_control_workspace_20240726/job_history/3336ee4b-c2df-4858-816a-4265eb735248

differences found
* what sites are in test that are not in truth
```
(base) ~/Downloads () $ bcftools isec -C test.vcf.gz truth.vcf.gz  | wc -l
Note: -w option not given, printing list of sites...
       0
```

* what sites are in the truth that are not in the test
```
(base) ~/Downloads () $ bcftools isec -C truth.vcf.gz test.vcf.gz | wc -l
Note: -w option not given, printing list of sites...
    6821
```
we stored these sites in truth_only_sites.txt

taking those sites and selecting them out of the truth vcf  we shold expect none of those sites to be imputed (has the annotation IMP in it) but that is not what we see
```
(base) ~/Downloads () $ bcftools view -R truth_only_sites.txt truth.vcf.gz | grep -v "#" | grep -i "imp" | wc -l
    1000
```

looking at these variants the large majority of the are imputed indels.  these indels "cover" the sites that were in our truth_only sites so they get added to the output from the previous command even though they show up in both the test and truth vcfs

For SNP sites, they followed the pattern of having two rows in the truth vcf.  One pointing at an imputed variant and one pointing at a non imputed variant (this is the one that is missing from the test vcf)

example variants
```
chr20	49224966	chr20:49224966:G:A	G	A	.	PASS	.	GT	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0	0/0
chr20	49224966	chr20:49224966:G:C	G	C	.	PASS	AF=0;DR2=0;IMP	GT:DS	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0	0|0:0
```

all of the above makes us pretty confident that these changes are only causing us to lose non imputed variants which are variants that were in the input that were not in the reference panel which we are ok with losing.

also a full genome run with the changes in the code - https://bvdp-saturn-dev.appspot.com/#workspaces/teaspoons-imputation-dev/teaspoons_imputation_dev_control_workspace_20240726/job_history/e727dcd8-0bab-455f-86c9-ae77f65170af

----

### Checklist 


kachulis (approved):
all looks good.

pipelines/broad/arrays/imputation_beagle/ImputationBeagle.wdl
```diff
@@ -84,42 +85,22 @@ workflow ImputationBeagle {
           gatk_docker = gatk_docker
       }
 
-      call tasks.CountVariantsInChunksBeagle {
+      call beagleTasks.CountVariantsInChunksBeagle {
```
mmorgantaylor, line 88: does it make sense to rename the tasks themselves to not include Beagle, since they're in the ImputationBeagleTasks wdl? or would that be more confusing because then there are two tasks with the same name (albeit coming from different source wdls)? I don't really have an opinion here

tasks/broad/ImputationTasks.wdl
```diff
@@ -1164,185 +1007,3 @@ task CreateVcfIndex {
     File vcf_index = "~{vcf_basename}.tbi"
   }
 }
-
-task PreSplitVcf {
```
mmorgantaylor, line 1168: just for my understanding, PreSplitVcf and PreChunkVcf were tasks we'd added to make this work nice on Azure, where we couldn't stream, and we've subsequently removed those tasks from the WDL so we can delete them here entirely? should we go ahead and delete [ImputationBeaglePreChunk.wdl](https://github.com/broadinstitute/warp/blob/TSPS-183_mma_beagle_imputation_hg38/pipelines/broad/arrays/imputation_beagle/ImputationBeaglePreChunk.wdl) since I think that does use these tasks?

mmorgantaylor (commented):
some minor comments, question about the ImputationBeaglePreChunk wdl

tasks/broad/ImputationTasks.wdl
jsotobroad: oh i totally forgot that wdl even existed.  I dont think we're ever gonna need that version and by the time we do we should have streaming so totally ok getting rid of that wdl

pipelines/broad/arrays/imputation_beagle/ImputationBeagle.wdl
jsotobroad: yea I like removing `Beagle` from the names

mmorgantaylor (approved):
thanks, this is awesome!!

PR #1501, opened 2025-02-06
Lk pd 2877 multiome peak calling

ekiernan:
This PR uses output from JoinBarcodes for Multiome Peak Calling.


website/docs/Pipelines/Multiome_Pipeline/README.md
```diff
@@ -82,7 +82,7 @@ Multiome can be deployed using [Cromwell](https://cromwell.readthedocs.io/en/sta
 | adapter_seq_read1 | Optional string describing the adapter sequence for ATAC read 1 paired-end reads to be used during adapter trimming with Cutadapt; default is "GTCTCGTGGGCTCGGAGATGTGTATAAGAGACAG". | String |
 | adapter_seq_read3 | Optional string describing the adapter sequence for ATAC read 2 paired-end reads to be used during adapter trimming with Cutadapt; default is "TCGTCGGCAGCGTCAGATGTGTATAAGAGACAG". | String |
 | run_cellbender | Optional boolean used to determine if the Optimus (GEX) pipeline should run CellBender on the output gene expression h5ad file, `h5ad_output_file_gex`; default is "false".  | Boolean |
-| run_peak_calling | Optional boolean used to determine if the ATAC pipeline should run Peak Calling; default is "false".  | Boolean |
+| run_peak_calling | Optional boolean used to determine if the ATAC pipeline should run Peak Calling; default is `false`. When set to true, the pipeline takes the ATAC h5ad produced by the JoinBarcodes task and performs peak calling to produce a cellbybin matrix and a cellbypeak matrix. | Boolean |
```
nikellepetrillo, line 85: should this be "cell by bin" and "cell by peak"? We are not using "cellbybin" below on lines 139/140

nikellepetrillo (approved):
small comment, thanks liz!

pipelines/skylab/multiome/Multiome.wdl
```diff
@@ -124,7 +125,8 @@ workflow Multiome {
             atac_nhash_id = atac_nhash_id,
             adapter_seq_read3 = adapter_seq_read3,
             atac_expected_cells = expected_cells,
-            peak_calling = run_peak_calling
```
jessicaway, line 127: why not use the top level input here?

jessicaway (approved):
One question, but looks good overall

pipelines/skylab/multiome/Multiome.wdl
ekiernan: The ATAC workflow has peak calling built in, so if we set it to top level here, it will run at this step. But we don't want the peak calling to run yet. We want to run peak calling as a separate task after JoinBarcodes so that we can use the JoinBarcodes output.

PR #1618, opened 2025-06-13
simplify import ../../tasks/broad down to nothing in tasks/broad

fredlas:
### Description

Barring a quirk I'm not aware of, these import paths are needlessly clunky, presumably after some large scale directory structure reshuffling.

----

### Checklist 

- [x] Did you modify, delete or move: file paths, file names, input names, output names, or task names?


fredlas:
Hmm, looks like maybe there is in fact such a quirk? Probably not worth any significant effort; feel free to close.

PR #1807, opened 2026-04-02
Lk aou mito post processing

ekiernan:
This PR adds a WDL for running aou mitochondria post processing, which includes generating the QC plots and additional filtering to produce a vcf with numt false positive warning as well as sample metadata with coverage metrics, contamination metrics and haplogroups.


rsc3 (approved):

verification/test-wdls/scripts/test_mito_post_processing_wdl_hail_logic.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env python3
```
jessicaway: Should this live next to the code it's testing? I don't think we have a convention in Warp let alone in AoU in Warp

jessicaway (approved):
Looks good 1 minor comment

verification/test-wdls/scripts/test_mito_post_processing_wdl_hail_logic.py
jessicaway: Actually, maybe this belongs next to the python code it tests in WarpTools?

PR #1811, opened 2026-04-11
fix: uses `Int` instead of `String` for `cpu` runtime values

claymcleod:
I've been working on getting `sprocket lint --hide-warnings --hide-notes` passing across the repository and fixing issues incrementally as I find them. This PR addresses one of those: several tasks declared their `cpu` runtime values as `String` when they should be `Int` (or `Float`).

While the WDL 1.0 spec left runtime keys loosely defined, the 1.1 spec formalized `cpu` with explicit type requirements:

> **`cpu`**
>
> * Accepted types:
>     * `Int`
>     * `Float`
> * Default value: `1`

(from the [WDL 1.1 specification](https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#cpu))

This doesn't affect execution on Cromwell, but it isn't correct per the WDL specification and `sprocket lint` flags it accordingly. I figure it's worth going ahead and updating.

### DIFF
--- a/all_of_us/PCA/pca_only_no_labels.wdl
+++ b/all_of_us/PCA/pca_only_no_labels.wdl
@@ -75,7 +75,7 @@ task ConcatenateChromosomalVcfs {
     runtime {
         docker: bcftools_docker
         memory: "${memory_gb} GB"
-        cpu: "${cpu}"
+        cpu: cpu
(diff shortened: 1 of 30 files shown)


jessicaway:

PR #1812, opened 2026-04-11
fix: replaces multi-option placeholders with function equivalents

claymcleod:
Continuing the `sprocket lint` cleanup, this fixes the cases where multiple placeholder options were combined in a single expression (e.g., `default` + `sep`, or a `true`/`false` guard paired with a `default` + `sep` placeholder on the same variable). These are replaced with the equivalent `prefix()`, `sep()`, and `select_first()` function calls.

Depends on #1811.


jessicaway:

PR #1646, opened 2025-08-11
wdl reorg Sonnet 4 take 2

rsc3:
### Description

https://docs.google.com/presentation/d/1giz5SEQGZXk_3pLoFxSysIhccZXlG0kEUAMwaxIhYoY/edit?usp=sharing

----

### Checklist 
If you can answer "yes" to the following items, please add a checkmark next to the appropriate checklist item(s) **and** notify our WARP team by tagging @broadinstitute/warp-admins in a comment on this PR.

- [ ] Did you add inputs, outputs, or tasks to a workflow?
- [ ] Did you modify, delete or move: file paths, file names, input names, output names, or task names?
- [ ] If you made a changelog update, did you update the pipeline version number?

PR #1813, opened 2026-04-11
fix: update broken relative imports in deprecated pipelines

claymcleod:
When files were moved into `deprecated/pipelines/` starting in 784c460f7 (#1541) and further reorganized in 973e83145 (#1646), the relative import paths weren't updated to account for the new directory depth. This fixes all 79 broken imports across 29 files.

Depends on #1812.


claymcleod:
I just realized these are all in the `deprecated/` folder, so rather than fixing them, I'll just exclude them from the `lint` run.

PR #1814, opened 2026-04-11
fix: quote integer literals in placeholder `default` options

claymcleod:
Placeholder options like `default` expect string values per the WDL spec. This quotes the bare integer literals (`default=0` to `default="0"`, `default=250` to `default="250"`) in `GermlineVariantDiscovery.wdl` and `UltimaGenomicsWholeGenomeGermlineTasks.wdl`.

Depends on #1812.


jessicaway:

PR #1815, opened 2026-04-11
fix: convert draft-2 WDL files in `all_of_us/rna_seq/GTEx/` to version 1.0

claymcleod:
Seven task/workflow files in `all_of_us/rna_seq/GTEx/` were still using draft-2 WDL syntax (no version statement, bare inputs outside of `input {}` blocks). This converts them to WDL 1.0 by adding `version 1.0` declarations and wrapping inputs in `input {}` blocks. The existing `${}` placeholder syntax and `command { }` blocks are preserved since they're still valid in 1.0.

Also changes the `cpu` runtime key from a string (e.g., `cpu: "${num_threads}"`) to a bare `Int` (e.g., `cpu: num_threads`) across all six task files, future-proofing for when runtime keys are formally typed (same strategy as #1811).

Depends on #1814.

### DIFF
--- a/all_of_us/PCA/pca_only_no_labels.wdl
+++ b/all_of_us/PCA/pca_only_no_labels.wdl
@@ -75,7 +75,7 @@ task ConcatenateChromosomalVcfs {
     runtime {
         docker: bcftools_docker
         memory: "${memory_gb} GB"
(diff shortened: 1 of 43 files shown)


jessicaway:

PR #1816, opened 2026-04-11
fix: resolve type mismatch errors across WDL files

claymcleod:
Fixes type mismatch errors flagged by `sprocket lint` across 5 files. These fall into a few categories:

- **Int-to-String coercion.** `range()` returns `Array[Int]`, but some task inputs expect `String id`. We use string interpolation (`"~{i}"`) to convert.
- **Conditional optional unwrapping.** The pattern `if defined(x) then x else y` doesn't narrow the type of `x` from `T?` to `T` in the `then` branch per the WDL spec. Replaced with `select_first([x, y])`.
- **Incorrect declared types.** `String java_memory_size = (memory_size - 1) * 1000` assigns an `Int` expression to a `String` — changed to `Int`. Similarly, `Array[Int] mem_gb_per_chunk = read_lines(...)` is invalid since `read_lines` always returns `Array[String]`.
- **Boolean in Map[String, String].** `InputQC.passes_qc` is `Boolean`, but the map literal expects `String` values. Wrapped with `"~{...}"`.
- **read_lines to Array[Int].** WDL has no `String` → `Int` coercion in any spec version (checked 1.0, 1.1, and 1.2). Changed `GetFingerprintingIntervalIndices` to write JSON and use `read_json()` so the output can be typed as `Array[Int]` for array indexing.

Note that there are remaining type mismatch errors related to passing optional types (`T?`) to non-optional call inputs that have defaults (`T x = default`). These are left as-is because the code is actually correct — passing `None` to a defaulted input means "use the default." This behavior was formalized in the WDL 1.2 spec (openwdl/wdl#462, openwdl/wdl#634) and is already implemented by Cromwell and miniwdl for all WDL versions. openwdl/wdl#761 tracks backporting this clarification to the 1.0 and 1.1 specs, after which Sprocket will implement it and these errors will resolve.

Depends on #1815.

### DIFF
--- a/all_of_us/ancestry/determine_hq_sites_intersection.wdl
+++ b/all_of_us/ancestry/determine_hq_sites_intersection.wdl
@@ -51,7 +51,7 @@ workflow determine_hq_sites_intersection {
                 vcf_intervals = training_vcf_so_bgz,
                 vcf_intervals_idx = training_vcf_so_bgz_idx,
                 intersecting_intervals=intersecting_intervals,
-                id = i
+                id = "~{i}"
         }
 
         call intersect_vcfs_as_sites_only {
@@ -92,7 +92,7 @@ workflow determine_hq_sites_intersection {
                 sites_only_vcf_idx = merge_sites_only_intersection.merged_vcf_idx,
                 output_name="full_data_sites_filtered",
                 service_account_json=service_account_json,
-                id = j
+                id = "~{j}"
         }
     }
 
@@ -135,7 +135,7 @@ task sitesOnlyAndHQFilterVcf {
     String output_filename = basename(vcf) + "." + id + ".sites_only.vcf.gz"
     String output_filename_idx = output_filename + ".tbi"
     String has_service_account_file = if (defined(service_account_json)) then 'true' else 'false'
-    String service_account_basename_pre = if (defined(service_account_json)) then service_account_json else ''
+    String service_account_basename_pre = select_first([service_account_json, ''])
     String service_account_basename = basename(service_account_basename_pre)
     String input_vcf_basename = basename(vcf)
     String updated_input_vcf = if (defined(service_account_json)) then input_vcf_basename else vcf
@@ -270,7 +270,7 @@ task filter_by_sites_only {
     String output_filename = basename(output_name) + "." + id + ".vcf.bgz"
     String output_filename_idx = output_filename + ".tbi"
     String has_service_account_file = if (defined(service_account_json)) then 'true' else 'false'
-    String service_account_basename_pre = if (defined(service_account_json)) then service_account_json else ''
+    String service_account_basename_pre = select_first([service_account_json, ''])
     String service_account_basename = basename(service_account_basename_pre)
     String input_vcf_basename = basename(vcf)
     String updated_input_vcf = if (defined(service_account_json)) then input_vcf_basename else vcf
--- a/all_of_us/ancestry/determine_hq_sites_intersection.wdl
+++ b/all_of_us/ancestry/determine_hq_sites_intersection.wdl
@@ -178,7 +178,7 @@ task sitesOnlyAndHQFilterVcf {
     runtime {
         docker:"us.gcr.io/broad-gatk/gatk:4.2.0.0"
         memory: "12 GB"
-        cpu: "4"
+        cpu: 4
         disks: "local-disk 100 HDD"
     }
 }
@@ -244,7 +244,7 @@ task merge_vcf_bgzs {
     runtime {
         docker: "mgibio/bcftools-cwl:1.12"
         memory: "120 GB"
-        cpu: "16"
+        cpu: 16
         disks: "local-disk 1500 HDD"
         bootDiskSizeGb: 1500
     }
@@ -303,7 +303,7 @@ task filter_by_sites_only {
     runtime {
         docker:"us.gcr.io/broad-gatk/gatk:4.2.0.0"
         memory: "7 GB"
-        cpu: "1"
+        cpu: 1
         disks: "local-disk 100 HDD"
     }
 }
@@ -340,7 +340,7 @@ task intersect_vcfs_as_sites_only {
     runtime {
         docker: "us.gcr.io/broad-gatk/gatk:4.2.0.0"
         memory: "7 GB"
-        cpu: "1"
+        cpu: 1
         disks: "local-disk 500 HDD"
     }
 }
--- a/pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.wdl
+++ b/pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.wdl
@@ -288,7 +288,7 @@ task ComputeShardsAndMemoryPerShard {
 
     output {
         Array[String] reference_chunk_file_paths = read_lines("reference_shard_file_paths.tsv")
-        Array[Int] mem_gb_per_chunk = read_lines("memory_per_chunk.tsv")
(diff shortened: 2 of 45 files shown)


all_of_us/ancestry/determine_hq_sites_intersection.wdl
```diff
@@ -51,7 +51,7 @@ workflow determine_hq_sites_intersection {
                 vcf_intervals = training_vcf_so_bgz,
                 vcf_intervals_idx = training_vcf_so_bgz_idx,
                 intersecting_intervals=intersecting_intervals,
-                id = i
+                id = "~{i}"
```
claymcleod, line 54: `range()` returns `Array[Int]`, but the `sitesOnlyAndHQFilterVcf` task declares `String id`. WDL doesn't coerce `Int` to `String` implicitly, so we use string interpolation to convert.

all_of_us/ancestry/determine_hq_sites_intersection.wdl
```diff
@@ -92,7 +92,7 @@ workflow determine_hq_sites_intersection {
                 sites_only_vcf_idx = merge_sites_only_intersection.merged_vcf_idx,
                 output_name="full_data_sites_filtered",
                 service_account_json=service_account_json,
-                id = j
+                id = "~{j}"
```
claymcleod, line 95: Same `Int` to `String` fix as above — `j` comes from `range()` but `filter_by_sites_only` expects `String id`.

all_of_us/ancestry/determine_hq_sites_intersection.wdl
```diff
@@ -135,7 +135,7 @@ task sitesOnlyAndHQFilterVcf {
     String output_filename = basename(vcf) + "." + id + ".sites_only.vcf.gz"
     String output_filename_idx = output_filename + ".tbi"
     String has_service_account_file = if (defined(service_account_json)) then 'true' else 'false'
-    String service_account_basename_pre = if (defined(service_account_json)) then service_account_json else ''
+    String service_account_basename_pre = select_first([service_account_json, ''])
```
claymcleod, line 138: The pattern `if defined(x) then x else y` doesn't narrow `x` from `String?` to `String` in the `then` branch per the WDL spec, so Sprocket flags the result as `String?` being assigned to `String`. `select_first([x, y])` does the same thing and returns the correct non-optional type.

all_of_us/ancestry/determine_hq_sites_intersection.wdl
```diff
@@ -270,7 +270,7 @@ task filter_by_sites_only {
     String output_filename = basename(output_name) + "." + id + ".vcf.bgz"
     String output_filename_idx = output_filename + ".tbi"
     String has_service_account_file = if (defined(service_account_json)) then 'true' else 'false'
-    String service_account_basename_pre = if (defined(service_account_json)) then service_account_json else ''
+    String service_account_basename_pre = select_first([service_account_json, ''])
```
claymcleod, line 273: Same `select_first` fix as in `sitesOnlyAndHQFilterVcf` above — identical pattern in the `filter_by_sites_only` task.

claymcleod (commented):
Inline comments explaining each fix for reviewer context.

pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.wdl
```diff
@@ -288,7 +288,7 @@ task ComputeShardsAndMemoryPerShard {
 
     output {
         Array[String] reference_chunk_file_paths = read_lines("reference_shard_file_paths.tsv")
-        Array[Int] mem_gb_per_chunk = read_lines("memory_per_chunk.tsv")
+        Array[String] mem_gb_per_chunk = read_lines("memory_per_chunk.tsv")
```
claymcleod, line 291: `read_lines()` always returns `Array[String]` per the WDL spec. This output isn't used anywhere in the workflow currently, so the type change is safe. If it's consumed downstream in the future, the values are integer strings from `int(np.ceil(x))` in the Python command.

tasks/wdl/JointGenotypingTasks.wdl
```diff
@@ -897,7 +897,7 @@ task CrossCheckFingerprint {
   Int cpu = if num_gvcfs < 32 then num_gvcfs else 32
   # Compute memory to use based on the CPU count, following the pattern of
   # 3.75GiB / cpu used by GCP's pricing: https://cloud.google.com/compute/pricing
-  Int memory = if defined(machine_mem_mb) then machine_mem_mb else round(cpu * 3.75 * 1024)
+  Int memory = select_first([machine_mem_mb, round(cpu * 3.75 * 1024)])
```
claymcleod, line 900: `machine_mem_mb` is `Int?` (line 886). The `if defined(x) then x else y` pattern doesn't narrow the optional, so the result type is `Int?` rather than `Int`. `select_first` does the same thing with correct typing.

tasks/wdl/JointGenotypingTasks.wdl
```diff
@@ -1057,10 +1057,18 @@ task GetFingerprintingIntervalIndices {
     else
       touch indices.out
     fi
+
+    python3 -c "
+with open('indices.out') as f:
+    indices = [int(line.strip()) for line in f if line.strip()]
```
claymcleod, line 1063: The original output used `read_lines()` which returns `Array[String]`, but callers (`JointGenotyping.wdl:397`, `UltimaGenomicsJointGenotyping.wdl:269`) need `Array[Int]` for array indexing. WDL has no `String` → `Int` coercion in any spec version, so we convert the indices file to JSON and use `read_json()` which can deserialize directly to `Array[Int]`.

tasks/wdl/Qc.wdl
```diff
@@ -503,7 +503,7 @@ task CollectRawWgsMetrics {
   Int disk_size = ceil(size(input_bam, "GiB") + ref_size) + additional_disk
 
   Int memory_size = ceil((if (disk_size < 110) then 5 else 7) * memory_multiplier)
-  String java_memory_size = (memory_size - 1) * 1000
+  Int java_memory_size = (memory_size - 1) * 1000
```
claymcleod, line 506: The RHS `(memory_size - 1) * 1000` is `Int` arithmetic, but the declaration was `String`. This worked at runtime because the value gets interpolated into the command string anyway, but the declared type should match the expression type.

verification/test-wdls/TestArrayImputationQC.wdl
```diff
@@ -49,7 +49,7 @@ workflow TestArrayImputationQC {
     call WriteMapToTsv {
       input:
         input_map = {
-          "passes_qc": InputQC.passes_qc,
+          "passes_qc": "~{InputQC.passes_qc}",
```
claymcleod, line 52: `InputQC.passes_qc` is `Boolean`, but this is a `Map[String, String]` literal so all values must be `String`. Wrapping in `"~{...}"` converts the boolean to its string representation ("true" or "false").

tasks/wdl/JointGenotypingTasks.wdl
claymcleod: This is a little more involved—I have not tested it myself, but perhaps someone should.

jessicaway:

claymcleod:
Thanks @jessicaway. It would be great if these workflows could be used more generally, please let me know if there's anything I can do to help.

rsc3:
Hello,

If you could remove (or start a new PR) any of these changes in the AoU directory here:

all_of_us/**

then we can go ahead and merge the changes across all of the (non-beta) pipelines outside this directory.

Thank you!

PR #1818, opened 2026-04-14
Rc 3251 scanvi

rsc3:
### Description

New pipeline for scANVI with terra GPU support

----

### Checklist 

### DIFF
--- a/pipeline_versions.txt
+++ b/pipeline_versions.txt
@@ -29,3 +29,4 @@ WholeGenomeReprocessing	 3.3.7	2026-01-21
 atac	 2.9.3	2025-09-19 
 scANVI	 1.0.1	2026-04-14 
 snm3C	 4.1.1	2025-09-19 
+test_cuda	 1.0.0	2026-04-14 
--- a/pipeline_versions.txt
+++ b/pipeline_versions.txt
@@ -27,4 +27,5 @@ VariantCalling	 2.2.8	2026-01-21
 WholeGenomeGermlineSingleSample	 3.3.7	2026-01-21 
 WholeGenomeReprocessing	 3.3.7	2026-01-21 
 atac	 2.9.3	2025-09-19 
+scANVI	 1.0.1	2026-04-14 
 snm3C	 4.1.1	2025-09-19 
--- a/pipelines/wdl/scanvi/README.md
+++ b/pipelines/wdl/scanvi/README.md
@@ -110,21 +110,46 @@ Instead, the pipeline:
 
 This ensures zero duplication: every preprocessing step executes exactly once (on CPU), and the GPU node spends 100 % of its time on model training and inference.
 
+## Runtime Configuration
+
+Both tasks use the same Docker image (pinned by digest). GPU and CUDA setup is handled entirely by the execution engine — the container does not configure the GPU environment itself.
+
+### Task 1 — `PreprocessFilter` (CPU-only)
+
+| Attribute | Value |
+|---|---|
+| `docker` | `us.gcr.io/broad-gotc-prod/scvi-scanvi@sha256:81fe915a045bd2929a1c457f4a0061055c6ea42fa3f88e9352b618e4a6e47b58` |
+| `bootDiskSizeGb` | 20 |
+| `disks` | `local-disk 1000 SSD` |
+| `memory` | `120 GiB` |
+| `cpu` | 32 |
+| `maxRetries` | 1 |
+
+### Task 2 — `MultiomeLabelTransfer` (GPU)
+
+| Attribute | Value |
+|---|---|
+| `docker` | `us.gcr.io/broad-gotc-prod/scvi-scanvi@sha256:81fe915a045bd2929a1c457f4a0061055c6ea42fa3f88e9352b618e4a6e47b58` |
+| `bootDiskSizeGb` | 20 |
+| `disks` | `local-disk 500 SSD` |
+| `memory` | `120 GiB` |
+| `cpu` | 32 |
+| `gpuType` | `nvidia-tesla-t4` |
+| `gpuCount` | 2 |
+| `nvidiaDriverVersion` | `525.147.05` |
+| `zones` | `us-central1-a us-central1-c` |
+| `maxRetries` | 1 |
+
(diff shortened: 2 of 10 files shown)


pipelines/wdl/scanvi/scANVI.wdl
```diff
@@ -0,0 +1,400 @@
...
+        fi
+
+        export GEX_FILE ATAC_FILE REF_FILE
+
+        # Symlink the gene annotation file (needed by snapatac2 make_gene_matrix)
+        ln -sf /usr/local/gencode.v41.basic.annotation.gff3.gz .
+
+        # ── Run preprocessing in Python ───────────────────────────────────────
+        python3 <<'PYEOF'
+import os
```
ekiernan, line 148: Might want to update formatting here, it looks like these lines didn't get indented

pipelines/wdl/scanvi/scANVI.wdl
```diff
@@ -0,0 +1,400 @@
...
+        cpu: nthreads
+        gpuType: "nvidia-tesla-t4"
+        gpuCount: 2
+        nvidiaDriverVersion: "525.147.05"
+        zones: "us-central1-a us-central1-c"
+        maxRetries: 1
+    }
+
+    output {
+        File scanvi_predictions_h5ad = "SCANVI_predictions.h5ad"
```
ekiernan: A lot of our pipelines take in some sort of input_id so that outputs can be named according to their specific input. We might consider adding now or at a future iteration.

pipelines/wdl/scanvi/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+The container script `multiome_label_transfer.py` bundles preprocessing **and** model training inside a single `main()` function. If the GPU task called `main()`, all preprocessing would run a second time on the GPU node — wasting expensive GPU hours on CPU-bound work that has already been completed.
+
+Instead, the pipeline:
+- Runs all preprocessing in **PreprocessFilter** on a CPU-only VM (no GPU cost).
+- In **MultiomeLabelTransfer**, imports only the three model-training / label-transfer / finalization functions from the script, bypassing `main()` entirely.
+
+This ensures zero duplication: every preprocessing step executes exactly once (on CPU), and the GPU node spends 100 % of its time on model training and inference.
+
+## Requirements
```
ekiernan, line 145: Do we want to add anything about the specific gpuType in the runtime or the nvidiaDriverVersion parameter for context as to why we are using those?

pipelines/wdl/scanvi/scANVI.wdl
rsc3: Its correct, unfortunate use of python and bash

pipelines/wdl/scanvi/scANVI.wdl
rsc3: When adding scatter support would make sense.

pipelines/wdl/scanvi/README.md
rsc3: Added the runtime requirements to the readme as working with Terra

ekiernan (approved):

pipelines/wdl/scanvi/scANVI.changelog.md
```diff
@@ -0,0 +1,16 @@
+# 1.0.1
```
jessicaway, line 1: Given that we did not release 1.0.0, does it make sense to collapse the changelog to only include a 1.0.0 version? We can keep the detail if it it helpful, but maybe just one version

pipeline_versions.txt
```diff
@@ -27,4 +27,6 @@ VariantCalling	 2.2.8	2026-01-21
 WholeGenomeGermlineSingleSample	 3.3.7	2026-01-21 
 WholeGenomeReprocessing	 3.3.7	2026-01-21 
 atac	 2.9.3	2025-09-19 
+scANVI	 1.0.1	2026-04-14
```
jessicaway, line 31: This will also change if we switch to 1.0.0 for scANVI

pipelines/wdl/scanvi/scANVI.wdl
```diff
@@ -0,0 +1,410 @@
...
+
+        export GEX_FILE ATAC_FILE REF_FILE
+
+        # Symlink the gene annotation file (needed by snapatac2 make_gene_matrix)
+        ln -sf /usr/local/gencode.v41.basic.annotation.gff3.gz .
+
+        # ── Run preprocessing in Python ───────────────────────────────────────
+        python3 <<'PYEOF'
+import os
+import anndata as ad
```
jessicaway, line 149: I know Liz already pointed out the indentation, but it does look strange to me too. Most of our in-line python has indentation that looks more consistent, might be the difference between using the CODE designation vs PYEOF? Doesn't need to block this PR, but might be worth looking into

pipelines/wdl/scanvi/scANVI.wdl
```diff
@@ -0,0 +1,410 @@
...
+workflow scANVI {
+
+  meta {
+    description: "Pipeline for cell type label transfer on Multiome data using SCVI and SCANVI models. Integrates single-cell RNA (GEX) and ATAC data with an annotated reference to transfer cell type labels via semi-supervised deep generative models."
+    allowNestedInputs: true
+  }
+
+  input {
+      # GCS bucket path containing input h5ad files (e.g., gs://bucket/path/to/inputs)
+      String? input_bucket
```
jessicaway, line 12: Since there are two possible input methods and both are optional, should we add checks to make sure the user has provided the necessary files in at least one way? I assume the pipeline will fail if you don't include the necessary inputs, but it might fail in some weird way with a cryptic error message

pipelines/wdl/scanvi/scANVI.wdl
```diff
@@ -0,0 +1,410 @@
...
+        run_id: "Unique identifier prepended to all output filenames."
+        docker: "Docker image containing the scvi-scanvi runtime environment."
+        disk_size: "Disk size in GB."
+        mem_size: "Memory size in GB."
+    }
+
+    command <<<
+        set -euo pipefail
+
+        # ── Resolve input file paths ──────────────────────────────────────────
```
jessicaway, line 118: maybe we can just add error messages here in case anything is missing?

jessicaway (approved):
Looks great! A couple minor comments

PR #1833, opened 2026-05-08
[TSPS-796] Improve memory footprint of RecomputeAndAnnotate, use correct MergeCoverageMetrics

MatthewBemis:
### Description

https://broadworkbench.atlassian.net/browse/TSPS-796

Some additional updates to the batched Glimpse2LowPassImputation WDL.

Previously the RecomputeAndAnnotate task used way too much memory. Even at 50 samples it needed a 32GB machine. The new version runs fine on 8.

Additionally the wrong task was used for merging the coverage metrics from the batches. That's been fixed to use the one from palantir-workflows.

----

### Checklist 


tasks/wdl/Glimpse2LowPassImputationTasks.wdl
```diff
@@ -124,37 +124,55 @@ task RecomputeAndAnnotate {
...
-        Int disk_size_gb = ceil(2.2 * size(merged_vcf, "GiB") + 50)
-        Int mem_gb
+        Int disk_size_gb = ceil(2.2 * size(merged_vcf, "GiB") + size(annotations, "GiB") + 50)
+        Int mem_gb = 8
         Int cpu = 2
         Int preemptible = 1
     }
 
     command <<<
         cat <<EOF > script.py
```
MatthewBemis, line 135: I got help from the robots for this re-write.

I did a before/after comparison of the outputs locally, and the recomputed AF and INFO values were all identical when accounting for floating point noise (i.e. no values were more than 0.00000000001 off)

tasks/wdl/Glimpse2LowPassImputationTasks.wdl
```diff
@@ -124,37 +124,55 @@ task RecomputeAndAnnotate {
...
-                    (2 * sum(num_samples) * aggregated_af * (1 - aggregated_af))
-
-annotation_dfs = [pd.read_csv(input_filename, sep='\t').rename(columns={'AF': f'AF_{i}', 'INFO': f'INFO_{i}'}) for i, input_filename in enumerate(input_filenames)]
-annotations_merged = functools.reduce(lambda left, right: pd.merge(left, right, on=['CHROM', 'POS', 'REF', 'ALT'], how='inner', validate='one_to_one'), annotation_dfs)
-annotations_merged['AF'] = annotations_merged.apply(lambda row: calculate_af(row), axis=1)
-annotations_merged['INFO'] = annotations_merged.apply(lambda row: calculate_info(row), axis=1)
-annotations_merged.to_csv('aggregated_annotations.tsv', sep='\t', columns=['CHROM', 'POS', 'REF', 'ALT', 'AF', 'INFO'], header=False, index=False)
+total_samples = sum(num_samples)
+num_batches = len(input_filenames)
+chunk_size = 100_000
```
jsotobroad: can we make this a task input with a default

tasks/wdl/Glimpse2LowPassImputationTasks.wdl
jsotobroad: what is the level of precision this is outputting?  We should match the precision that the glimpse tool outputs when i think is 4 digits?  More digits will make the outputs way larger.  

Did we already do an analysis of the difference in annotation values between batched output, non batched output, previous version of hte wdl?

MatthewBemis: first Q: I just checked the vcf output by glimpse and the highest level of precision i found was `0.000999615` for the INFO score and `0.999` for AF. so i can update this task to match that?

second Q: yeah i have done some analysis of this and there are some differences between the batched versus previous version of the wdl. i still need to reach a conclusion on whether or not they are tolerable or not. that will be much easier now that this wdl has stabilized and we have the non-determinism fixed

jsotobroad: gotcha, yea the recomputed annotations should match the precision of the stuff coming out of glimpse.  I'm **super** surprised glimpse returns something with that much precision for `INFO`....  that seems like such a waste.  I wonder if Chris would be ok with us limiting that precision to 3 digits or w/e like `AF` 🤔

pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.wdl
jsotobroad, line 85: I think Morgan mentioned this on a previous PR but do we want to skip tasks related to recomputing if there is only one batch? If there is only one batch we can go straight to selecting variant and hom ref records.

tasks/wdl/Glimpse2LowPassImputationTasks.wdl
MatthewBemis: I pinged Chris about this- if we get a quick 'do it' answer I can roll it into this PR but if not I'll do it as a quick followup PR so we can get this more memory-efficient implementation merged for all to use

pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.wdl
```diff
@@ -105,27 +104,31 @@ workflow Glimpse2LowPassImputation {
...
-            input:
-                merged_vcf = MergeContigVcfs.output_vcf,
-                annotations = ExtractAnnotations.annotations,
-                num_samples = batch_sample_count,
-                output_basename = output_basename + "." + contigs[contig_idx] + ".imputed.merged.reannotated",
-                docker_merge = docker_merge,
-                mem_gb = mem_gb_merge
+        # Recompute AF and INFO as weighted averages across batches and apply back to the merged VCF.
+        # Skip when there is only one batch since the single-batch annotations are already correct
+        if (length(SplitIntoSampleBatches.crams_batches) > 1) {
```
jsotobroad: i think we can encompass more tasks in this block.  like im not sure there is a need to run `ExtractAnnotations` or `MergeContigVcfs` in this case.

pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.changelog.md
```diff
@@ -1,3 +1,9 @@
+# 0.0.11
+2026-05-08 (Date of Last Commit)
```
jsotobroad: can you update the date now that other stuff has been merged in previous versions

pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.wdl
```diff
@@ -87,45 +86,48 @@ workflow Glimpse2LowPassImputation {
...
-        call Glimpse2LowPassImputationTasks.RecomputeAndAnnotate {
-            input:
-                merged_vcf = MergeContigVcfs.output_vcf,
-                annotations = ExtractAnnotations.annotations,
-                num_samples = batch_sample_count,
-                output_basename = output_basename + "." + contigs[contig_idx] + ".imputed.merged.reannotated",
-                docker_merge = docker_merge,
-                mem_gb = mem_gb_merge
-        }
+        File annotated_contig_vcf = select_first([RecomputeAndAnnotate.merged_imputed_vcf, batch_vcfs_for_contig[0]])
```
jsotobroad, line 119: love it!

jsotobroad (approved):
awesome this looks good. just one comment about dates.

Do we still need to do an analysis on the annotation differences between batching vs no batching?

MatthewBemis:
@jsotobroad yes- there's one final pair of submissions running now to compare the batched vs nonbatched outputs of this workflow

mmorgantaylor (approved):
thank you, this is great!

PR #1854, opened 2026-05-22
MT Merge - adjust median coverage calculation and change default data type to avoid overflow

jessicaway:
### Description

This PR updates the per-position median coverage calculation to match the way median is computed in Hail. For positions with an even number of samples, the code now uses the integer average of the two middle values rather than selecting the lower of the two middle values. In practice, this change did not alter any median values in our 50k-sample validation dataset, so we expect minimal downstream impact.

This PR also changes the default datatype used to store coverage values in the coverage HDF5 file from uint16 to uint32. In addition, we added an explicit overflow check that will fail if a value exceeds the representable range. This is expected to be extremely unlikely, since uint32 supports values up to approximately 4 billion. This increased the size of our h5 file from 1.53 GB to 1.92 GB for our 50k sample validation dataset. 

Corresponding Wrap-tools PR is here: https://github.com/broadinstitute/warp-tools/pull/197


ekiernan (approved):
ty!

PR #1886, opened 2026-07-13
TSPS-1061 Run SV imputation deterministically

mmorgantaylor:
### Description

A couple changes to allow setting the seed and number of threads in the glimpse phase task, to aid in running the wdl deterministically.

Glimpse phase cpu still defaults to 4, but setting `glimpse_phase_cpu_override` to 1 results in 1 thread in glimpse task, and this should be used when the pipeline needs to be run deterministically, i.e. in warp tests.

Setting cpu override to 1 results in 1 thread in glimpse task - here's the log from glimpse task:
```
* #Threads             : [1]
```

And the outputs of two such runs produced identical VCFs (aside from header). Tests linked in jira ticket.

Jira ticket: https://broadworkbench.atlassian.net/browse/TSPS-1061

----

### Checklist 

- [x] Did you add inputs, outputs, or tasks to a workflow?


pipelines/wdl/glimpse/sv_imputation/Glimpse2SVImputationBatch.wdl
```diff
@@ -161,7 +168,9 @@ task GLIMPSE2Phase {
         String output_region
         File genetic_map
         String output_prefix
-        String? extra_phase_args = "--thread $(nproc) --impute-reference-only-variants --keep-monomorphic-ref-sites --Kpbwt 1000 --main 10 --burnin 5 --err-imp 1E-3"
+        Int seed = 15052011
+        Int threads = 4
```
jsotobroad, line 173: So right now the value that is passed to the command and the value we have for runtime arent related and i think they should be.  so instead of passing in to the task
```
                runtime_attr_override = object {
                    cpu_cores: glimpse_phase_cpu_override
                },
```
I think you can put this `threads` input in the `default_attr` specified below.  since this is going to come up when we scatter this wdl by chromosome you could copy [what we did with low pass](https://github.com/broadinstitute/warp/blob/develop/pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputationBatch.wdl#L39) to pass in the `threads` override value as well

mmorgantaylor: ah that makes a lot more sense, thanks

pipelines/wdl/glimpse/sv_imputation/Glimpse2SVImputationBatch.wdl
```diff
@@ -161,7 +169,9 @@ task GLIMPSE2Phase {
         String output_region
         File genetic_map
         String output_prefix
-        String? extra_phase_args = "--thread $(nproc) --impute-reference-only-variants --keep-monomorphic-ref-sites --Kpbwt 1000 --main 10 --burnin 5 --err-imp 1E-3"
+        Int seed = 15052011
```
salonishah11, line 172: Out of curiosity, is there a reasoning behind this specific seed number?

salonishah11 (approved):

pipelines/wdl/glimpse/sv_imputation/Glimpse2SVImputationBatch.wdl
jsotobroad: its actually the [Glimpse sets as default](https://odelaneau.github.io/GLIMPSE/docs/documentation/phase/#basic-options) but just in case they change that its nice for ours to be consistent

salonishah11: Got it, thank you!

jsotobroad (approved):
nice!

PR #1891, opened 2026-07-17
TSPS-1082 Low Pass WGS Imputation Input QC: add cram-crai basename check

mmorgantaylor:
### Description

We expect users to submit cram-crai pairs that share a basename. If they don't, most likely the user has made a mistake. So we're adding this check to our InputQC wdl.

Test case output:
```
{ 
   "qc_messages": "Found 1 CRAM-CRAI pair with mismatched basenames: gs://pd-test-storage-public/Glimpse2LowPassImputationQC/input/plumbing/crams/fakeCram1.cram and gs://pd-test-storage-public/Glimpse2LowPassImputationQC/input/plumbing/crams/fakeCram2.cram.crai.", 
   "passes_qc": "false" 
}
```

Note we expect this new test case to fail because there's no truth data for it yet. Will create truth data once this PR is merged.

Also updated the crai required suffix from `.crai` to `.cram.crai` since we are effectively checking for that with the above check. This will likely cause a couple more tests to fail because the error message has changed.

Jira ticket: https://broadworkbench.atlassian.net/browse/TSPS-1082

----

### Checklist 

### DIFF
--- a/pipelines/wdl/glimpse/low_pass_imputation/input_qc/Glimpse2LowPassImputationQC.wdl
+++ b/pipelines/wdl/glimpse/low_pass_imputation/input_qc/Glimpse2LowPassImputationQC.wdl
@@ -215,11 +215,11 @@ task ValidateCramsAndIndicesAndSampleIds {
         else:
             print("All CRAM files have the correct .cram extension.")
 
-        cram_indices_with_wrong_extension = [c for c in cram_indices if not c.endswith('.crai')]
+        cram_indices_with_wrong_extension = [c for c in cram_indices if not c.endswith('.cram.crai')]
         if cram_indices_with_wrong_extension:
-            qc_messages.append(create_error_message_with_item_list(f"Found {pluralize(len(cram_indices_with_wrong_extension), 'CRAM index file')} without a .crai extension", cram_indices_with_wrong_extension))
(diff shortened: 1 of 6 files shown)


MatthewBemis (approved):
this is awesome- glad we're adding this check!

pipelines/wdl/glimpse/low_pass_imputation/input_qc/Glimpse2LowPassImputationQC.wdl
```diff
@@ -229,6 +229,18 @@ task ValidateCramsAndIndicesAndSampleIds {
         else:
             print("CRAM paths are unique.")
 
+        # Validate that each cram-crai pair has matching basenames
+        mismatched_basename_pairs = []
+        for cram, crai in zip(crams, cram_indices):
+            cram_basename = cram.split('/')[-1].replace('.cram', '')
+            crai_basename = crai.split('/')[-1].replace('.cram.crai', '')
```
jsotobroad: we probably want to make sure we are only replace the suffix for these with empty as opposed to replacing any occurence of the value in their file name.  You can probably replace this with https://www.geeksforgeeks.org/python/python-string-removesuffix/

jsotobroad (commented):
just one comment but looks good

jsotobroad (approved):

PR #1898, opened 2026-07-28
TSPS-1024 TSPS-1129 Make 50002 OOM less likely

jsotobroad:
### Description

We ran into a number of issues when scale testing / our friends and family testing with OOM tasks causing 50002s and not being retried with more memory which caused a high percentage of our workflows to fail.  These change will make the 50002s less likely to happen.

We also saw some out of disk failures after reducing the disk in the previous version so bumping that up to a higher number

JIRA tickets:
https://broadworkbench.atlassian.net/browse/TSPS-1129
https://broadworkbench.atlassian.net/browse/TSPS-1024

----

### Checklist 

### DIFF
--- a/pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.changelog.md
+++ b/pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.changelog.md
@@ -1,3 +1,8 @@
+# 1.0.3
+2026-07-28 (Date of Last Commit)
+
+* update docker image used for glimpse. This will include runtime optimizations
+
 # 1.0.2
 2026-07-20 (Date of Last Commit)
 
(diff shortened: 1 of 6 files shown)


pipelines/wdl/glimpse/low_pass_imputation/Glimpse2LowPassImputation.changelog.md
```diff
@@ -1,3 +1,9 @@
+# 1.0.3
+2026-07-28 (Date of Last Commit)
+
+* update docker image used for glimpse. This will include runtime optimizations
+* change deafult sample batch size to 500
```
mmorgantaylor:
```suggestion
* change default sample batch size to 500
```

mmorgantaylor (approved):
one typo fix, otherwise 👍

PR #1903, opened 2026-07-31
FIX scientific test failures and nondeterminisms that broke on recent staging to master

rsc3:
Notes below on each fix or why it was a transient error or known nondeterminism

Compare CrosscheckMetric order- and path-insensitively

Picard CompareMetrics diffs metric rows by index across every column. The CrosscheckMetric emits read-group rows in non-deterministic order and its LEFT_FILE/RIGHT_FILE columns hold absolute BAM paths (submission id + SortSampleBam attempt-N), so verification flags spurious diffs on runs where nothing scientific changed. Normalize (drop the two path columns, sort rows) and diff for this metric type only; all other metrics keep the exact Picard comparison. Real value drift is still caught.

# Multiome and Optimus nondeterminism tolerance fixes

# clean passes that GHA didn't pay attention too

RNA with UMIs
ReblockGVCF
UltimaGenomicsWholeGenomeGermline
UltimaGenomicsWholeGenomeCramOnly
UltimaGenomicsJointGenotyping
RNA with UMIs
ReblockGVCF

# Slide Seq 

Tolerate float-variance nondeterminism in gene metrics
  
  Problem
  
  TestSlideSeq (and, by the same mechanism, the other Optimus-family tests) fails on genomic_read_quality_variance. A single gene differs between the test run and truth:
  
  genomic_read_quality_variance:  test 3.86594677  vs  truth 3.865946531
  
  That's a 6.2e-8 relative difference on one gene.
  
  Root cause
  
  genomic_read_quality_variance (and its sibling genomic_read_quality_mean) is a per-gene floating-point aggregate. Its result depends on summation/accumulation order, so it drifts run-to-run at the last few significant digits — expected float nondeterminism, not a regression.
  
  These metrics were previously assumed deterministic and compared with no tolerance in two independent places:
  1. CompareGeneMetrics → CompareCompressedTextFiles, an exact whole-file diff <(gunzip | sort).
  2. CompareH5adFilesGEX → strict .equals() on each var column.
  
  Neither had a per-column tolerance hook, so a last-digit wobble in one gene fails the whole comparison.
  
  Change
  
  - Add CompareGeneMetricsWithTolerance: a wide-CSV compare keyed by the id column and order-insensitive. Every column is compared exactly, except a configurable tolerated_columns set (default genomic_read_quality_variance, genomic_read_quality_mean) which uses a 1e-4 relative tolerance. Non-numeric or exact-column
  differences still fail; gross drift still fails.
  - Swap it in for the gene-metrics comparison in VerifyOptimus, VerifyMultiome, VerifyPairedTag, VerifySlideSeq, VerifySlideTags. Cell and UMI metrics keep the exact diff — this only loosens the two named float columns.
  - Mirror the same 1e-4 relative tolerance in CompareH5adFilesGEX's post-var_names_make_unique gene loop (np.allclose(rtol=1e-4)), so the h5ad path agrees with the CSV path.
  
  Stdlib-only Python (csv/gzip) — no new dependencies.
  
  Scope / safety
  
  - Test-infrastructure only (verification/**); no pipeline WDL touched, so no pipeline_version bump or changelog.
  - A 1e-4 tolerance is ~1600× the observed drift while still catching any real change (a 0.9% injected drift fails).
  
  Testing
  
  - womtool validates VerifyTasks.wdl, all five swapped verifiers, and the Test* wrappers.
  - Unit-checked the comparator on synthetic gzip fixtures: 6e-8 variance drift → pass; exact-column change → fail; 0.9% variance drift → fail.
  
  Follow-up
  
  Reclassify genomic_read_quality_variance and genomic_read_quality_mean as nondeterministic in the WARP metrics catalog (both the cell-metric and gene-metric rows — they're currently marked "assumed deterministic"). The cell-metrics comparison still uses the exact diff; if a run ever drifts there, it's the same
  one-line swap to the tolerant task.
  


verification/VerifyMultiome.wdl
```diff
@@ -58,7 +58,8 @@ workflow VerifyMultiome {
         input:
             test_bam       = test_atac_bam,
             truth_bam      = truth_atac_bam,
-            lenient_header = true
+            lenient_header = true,
+            mappings_diff_threshold = 0.05
```
jessicaway, line 62: Is this threshold universally acceptable. Just want to make sure a human checks that this is a reasonable threshold for what to ignore

rsc3: I chose it, but happy to make it more restrictive. Note that besides the BAM we are also testing most of the cell, library, and gene metrics. So this test seems to be more about noticing any big differences in the BAMs.

ekiernan (approved):

jessicaway (approved):

khajoue2 (approved):

PR #1913, opened 2026-08-06
Cleanup dead wdl

rsc3:
Get rid of dead WDL files and unused imports

Some scANVI RRID updates


jessicaway (approved):
The difference between the two versions is MINIMAL. This probably happened with the test migration. Ok to delete the old and change the "new" to not say "updated". Thanks @rsc3

